PR #3, opened 2018-03-06
Models

KaiVolland:
**This introduces the basic setup for the models**

- Clean up some leftovers from `create-react-app`
- Adds styleguidist to package.json
- Introduces the `Rule` model


.editorconfig
```diff
@@ -0,0 +1,11 @@
+# EditorConfig is awesome: http://EditorConfig.org
+
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
```
marcjansen: … except for this one :grimacing:

package.json
```diff
@@ -27,13 +27,16 @@
...
+    "eject": "react-scripts-ts eject",
+    "styleguide": "styleguidist server",
+    "styleguide:build": "styleguidist build"
   },
   "devDependencies": {
     "@types/jest": "22.1.3",
     "@types/node": "9.4.6",
     "@types/react": "16.0.38",
     "@types/react-dom": "16.0.4",
+    "react-styleguidist": "^6.2.6",
```
marcjansen: Do we want the carret?

src/models/Rule/Rule.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+import Symbolizer from '../Symbolizer/Symbolizer';
+import Filter from '../Filter/Filter';
+import ScaleDenominator from '../ScaleDenominator/ScaleDenominator';
+
+it('is defined', () => {
+  expect(Rule).toBeDefined();
+});
+
+describe('Constructor', () => {
+  it('can be created with only without filter or scaledenominator', () => {
```
marcjansen: `it('cannot be created without symbolizer', () => {});` ?

src/models/Rule/Rule.spec.ts
```diff
@@ -0,0 +1,101 @@
...
+    const scaleDenominator = new ScaleDenominator();
+    const rule = new Rule(symbolizer, filter, scaleDenominator);
+    expect(rule).toBeDefined();
+    expect(rule.symbolizer).toBe(symbolizer);
+    expect(rule.filter).toBe(filter);
+    expect(rule.scaleDenominator).toBe(scaleDenominator);
+  });
+});
+
+describe('Accessors', () => {
```
marcjansen: Good that this is tested. Any chance of making this a less manual work? both when creating classes and testing them?

package.json
KaiVolland: As i'm pretty sure we want to use greenkeeper for the geostyler the carret should be removed. I take care of it.

src/models/Rule/Rule.spec.ts
KaiVolland: 1. Didn't found anything for my editor.
2. We should discuss if these properties need to be private anyway. If we use `getters` and `setters` they should do something. Otherwise we can just leave these properties public.
 I really don't want that JAVAish overhead but i was playing around with TypeScript features. :wink:

hwbllmnn:
Nice!

.editorconfig
```diff
@@ -0,0 +1,11 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+indent_style = space
+indent_size = 2
```
dnlkoch: Only minor, but I suggest to add:
```
charset = utf-8
trim_trailing_whitespace = true
```

src/models/Filter/Filter.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ *
+ *
+ * @class Rule
```
dnlkoch: `@class Rule` -> `@class Filter`

src/models/Rule/Rule.spec.ts
dnlkoch: Not related to the generation of test files, but this [extension](https://marketplace.visualstudio.com/items?itemName=DSKWRK.vscode-generate-getter-setter) looks promising for the work with typescript in combination with visual studio code.

src/models/ScaleDenominator/ScaleDenominator.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ *
+ *
+ * @class Rule
```
dnlkoch: `@class Rule` -> `@class ScaleDenominator`

src/models/Symbolizer/Symbolizer.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ *
+ *
+ * @class Rule
```
dnlkoch: `@class Rule` -> `@class Symbolizer`

PR #4, opened 2018-03-07
[WIP] Filter Interface & Operator Type

KaiVolland:


.vscode/tasks.json
```diff
@@ -0,0 +1,12 @@
+{
+    // See https://go.microsoft.com/fwlink/?LinkId=733558
+    // for the documentation about the tasks.json format
```
marcjansen: JSON does not have comments.

KaiVolland: ... this comes directly from VisualStudio Code.

marcjansen: ... I thought so. Not blocking.

src/Interface/Filter/Filter.ts
```diff
@@ -0,0 +1,14 @@
+import Operator from '../../Type/Operator';
+
+export interface Filter extends Array<any> {
+  0: Operator;
+}
+
+export interface Negation extends Filter {
```
chrismayer: Is it common in TS to have several ``interface`` definitions in one file?

chrismayer:
Looks good to me @KaiVolland. I posted a little question, which is more TypeScript related and which is non-blocking.

Furthermore: Don't you think it is cleaner to separate the changes in the ``tslint.json`` config from this PR?

src/Interface/Filter/Filter.ts
KaiVolland: Im not sure about that. I just did it as it was so little code. But i see the advantages of splitting these into mutliple files. I ll change this.

PR #10, opened 2018-04-17
Introduce models for "GeoStyler Data"

chrismayer:
This adds a first version for the models of the "GeoStyler Data" package.

Root object ``Data`` is an aggregation of the schema of the data (comparable to DescribeFeatureType) based on [JSON Schema](http://json-schema.org/) and the data itself (example features) as [GeoJSON](http://geojson.org/).


marcjansen:
Thanks!

marcjansen:
Is this still `Work in progress`?

chrismayer:
I expect some modifications since this can be seen as a draft for the models of the "GeoStyler Data" package. So I would appreciate feedback, which leads me to give this free for reviews.

chrismayer:
Thanks for your feedback @KaiVolland. I just addressed your remarks:
  - Using more ``type`` definitions instead of ``class``
  - Included a GeoJSON lib

Additionally:
  - Move the sources to the new folder ``Model``

Would you please have another look?

package.json
```diff
@@ -33,6 +33,7 @@
   },
   "devDependencies": {
     "ts-jest": "22.4.3",
+    "@types/geojson": "^7946.0.2",
```
KaiVolland: Please remove the `^`.

src/Type/DataSchema.ts
```diff
@@ -0,0 +1,53 @@
...
+ * @type DataSchema
+ */
+type DataSchema = {
+
+  /**
+   * Optional title for the described entity
+   *
+   * @type {string}
+   */
+  title: string | '';
```
KaiVolland: I think it should be `title?: string;` if its optional

src/Type/DataSchema.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  /**
+   * Optional type definition for the described entity
+   *
+   * @type {string}
+   */
+  type: string | 'object';
```
KaiVolland: I think it should be `type?: string;` if its optional. `'object'` allready is a string.

src/Type/DataSchema.ts
chrismayer: By this notation I wanted to ensure that ``title`` is always a String (= default value) and not ``null`` or ``undefined``.  Does this make sense?

src/Type/DataSchema.ts
chrismayer: By this notation I wanted to ensure that ``type`` is always a String (= default value) and not ``null`` or ``undefined``.  Does this make sense?

src/Type/DataSchema.ts
KaiVolland: I don't think it's possible to define default values for properties in types. I think your syntax would something like: title can be any string or the string ''...

src/Type/DataSchema.ts
KaiVolland: see above

src/Type/DataSchema.ts
KaiVolland: IMHO we should leave it as required for know. We can be more precise later on. So you may just remove the empty string thing.

chrismayer: Agree...

PR #12, opened 2018-04-17
Scaledenominator as a type

KaiVolland:
This refactors the ScaleDenominator to be a type instead of a class.

Also fixes some broken imports of the `Filter` interface.


KaiVolland:
IMHO we should always prefere the strictest definition with the least effort. Which means that i'd prefer Types over Interfaces over Classes.

I think the final target for the `GeoStyler_Data` and `GeoStyler_Style` will be TypeScript Type-Definitions distributed via http://definitelytyped.org/. As i'm not that deep into TypeScript yet this opinion might change in the course of time.

KaiVolland:
Thanks for review.

I'm gonna delve into https://www.typescriptlang.org/docs/handbook/declaration-files/introduction.html to find out how our declarations (not definitions --> allready learned something :smile: ) should look like.

Issue #32, opened 2018-04-20
Broken import in FilterUtil.spec.ts 


chrismayer:
In the file ``src/util/FilterUtil/FilterUtil.spec.ts`` seems to be a broken import definition for FilterUtil:

```
import { Filter } from '../../Interface/Filter/Filter';
```
Changing the import to 

```
import Filter from '../../Interface/Filter/Filter';
```
seems to solve this.

Issue #28, opened 2018-04-20
Implement DataProvider


chrismayer:
The GeoStyler needs a ``DataProvider`` component, which is supposed to do the following:

  - Retrieve and manage a list of available ``DataParser`` instances (for different formats)
  - Implement a method retrieving external geodata (e.g. GeoJSON), transform it into the internal data structure (by a suitable parser) and return internal data object

@KaiVolland and me agreed on the following structure (pseudo code):

```
new DataProvider({ 
  parser: [
    {
      format: 'GeoJSON',
      impl: new GeoJsonDataParser({ ... })
    }  
  ] 
});
```

PR #34, opened 2018-04-23
Implement DataProvider

chrismayer:
This adds a DataProvider component according to #28.

To resolve the dependency for ``geostyler-geojson-parser`` correctly this lets the package point to a special branch [with-dist-folder](https://github.com/terrestris/geostyler-geojson-parser/tree/with-dist-folder), which contains the compiled version of the GeoJSON parser. See https://github.com/terrestris/geostyler-geojson-parser/pull/5 for more information. 

The PR also contains a demo GUI to show the capabilities of the DataProvider.
**~~The demo UI is currently not runnable due to #32 (but can easily be fixed in local checkout).~~** 

image


chrismayer:
The failing tests are related to the fact that the package ``geostyler-geojson-parser`` points to this [github-branch](https://github.com/terrestris/geostyler-geojson-parser/tree/with-dist-folder). While using ``npm link`` in the local setup everything is fine. Maybe we have to publish the ``geostyler-geojson-parser`` via npm to make this runnable in a sustainable way.

src/DataProvider/DataProvider.ts
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  /**
+   * Imports the given data and transforms it into the internal data structure. Therefore an appropriate parser is
+   * detected by the given format.
+   * 
+   * @param inputData 
+   * @param format 
+   */
+  importData(inputData: any, format: string) {
```
KaiVolland: I would switch the parameter order. Just a feeling.

chrismayer:
I just added another commit, which includes the published packages for geostyler-data models
and the geostyler-geojson-parser from npmjs.org.

Since now Travis is green we might think about merging this PR.

chrismayer:
Thnaks @KaiVolland!

PR #64, opened 2018-05-16
Comparison Filter GUI

chrismayer:
This introduces a GUI visualizing a ComparisonFilter.  The GUI consists of a bunch of react components.

For demonstration purpose a little demo app has also been developed, where the following can be done:

- Select a GeoJSON file and import it into the app
- Select an attribute, an operator and a value to define the ComparisonFilter
- Dependent on the attribute type (String, Number, Boolean) a different UI is shown
- The internal ComparisonFilter object can be serialized and is printed into a text field 

image

UI for boolean attributes:
image


src/Component/Filter/AttributeCombo/AttributeCombo.tsx
```diff
@@ -0,0 +1,77 @@
...
+import { Select } from 'antd';
+import FormItem from 'antd/lib/form/FormItem';
+const Option = Select.Option;
+
+/**
+ * Combobox offering the attributes to be filtered on.
+ */
+class AttributeCombo extends React.Component<any, any> {
+
+  // PROPS
```
KaiVolland: React usually defines its props with [PropTypes](https://reactjs.org/docs/typechecking-with-proptypes.html).
As we make use of TypeScript [another approach with interfaces](https://charleslbryant.gitbooks.io/hello-react-and-typescript/content/TypeScriptAndReact.html) should be the way to go.

The props should not be declared as member variables. The declared member variables will not be associated with the props. So you will have `this.internalDataDef` and `this.props.internalDataDef` and so on.

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+          <Select style={{ width: '100%' }} onChange={this.props.onAttributeChange} placeholder={this.placeholder}>
+              {options}
+          </Select>
+
+        </FormItem>
+
+      </div>
+    );
+  }
```
KaiVolland: You might want to profit from object destructuring / spread operator for the props. You can then use "passThorughProps".
[Compare](https://github.com/terrestris/react-geo/blob/master/src/Field/CoordinateReferenceSystemCombo/CoordinateReferenceSystemCombo.jsx#L195)

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.css
```diff
@@ -0,0 +1,4 @@
+
```
KaiVolland: Please use less instead of css

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
```diff
@@ -0,0 +1,173 @@
...
+import './ComparisonFilterUi.css';
+
+/**
+ * UI for a ComparisonFilter consisting of
+ *
+ *   - A combo to select the attribute
+ *   - A combo to select the operator
+ *   - An input field for the value
+ */
+class ComparisonFilterUi extends React.Component<any, any> {
```
KaiVolland: Please remove "Ui" from the classname

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
```diff
@@ -0,0 +1,173 @@
...
+
+  // PROPS
```
KaiVolland: See comment above regarding props.

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
```diff
@@ -0,0 +1,173 @@
...
+      numberFieldVisible: false
+    };
+  }
+
+  /**
+   * Handler function, which is executed, when to underlying filter attribute changes.
+   *
+   * Changes the input field for the filter value and stores the appropriate attribute name as member.
+   */
+  onAttributeChange = (newAttrName: string) => {
```
KaiVolland: Please transform methods to following syntax and use `.bind(this)` if needed:

```tsx
onAttributeChange(newAttrName: string) {
```

Otherwise styleguidist doesn't recognize this methods as methods but as properties.

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
```diff
@@ -0,0 +1,77 @@
+import * as React from 'react';
```
KaiVolland: I'm not sure if it makes a difference but i'd prefer `import React from 'react';`.

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
```diff
@@ -0,0 +1,77 @@
...
+  /** label  for this combobox */
+  label: string = 'Attribute';
+
+  /** placeholder shown when combobox has no selection */
+  placeholder: string = 'Select Attribute';
+
+  /** Handler function bound to the 'onChange' of the underlying Select */
+  onAttributeChange: Function;
+
+  constructor(props: any) {
```
KaiVolland: If you don't perform any actrions with props or the state you can omit the constructor.

KaiVolland:
Looks very promising. :heart_eyes:  I added some notes to structural things.
Would be cool if you could add a screenshot once your happy with the visualization.

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
chrismayer: Good point, I have this on my list. This is the next thing I will tackle.

chrismayer:
Thanks for your review @KaiVolland! Good points, I agree with most of them. I will go on and address your remarks / hints.

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
chrismayer: Agree...

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.css
chrismayer: OK...

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
chrismayer: OK...

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
chrismayer: I will check that

src/Component/Filter/AttributeCombo/AttributeCombo.tsx
chrismayer: Does not work for me. The module doesn't seem to have a standard export. And using ``import { React } from 'react';`` also does not work because there is no exported member ``React`` in the module. 
I also just discovered `` import * as React from 'react'; `` in all documentations / tutorials I read so far.

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.tsx
KaiVolland: I wasn't aware that TypeScript recommends this way of declaring class methods. So 
we should stick to the TypeScript recommendation for know.

src/Component/Filter/ComparisonFilter/ComparisonFilterUi.css
chrismayer: I created a separate issue for this (#65), so we can tackle this later (since this seems a bit more complicated that one might think at first sight).

chrismayer:
I addressed or at least commented your remarks, @KaiVolland. I think this is now ready for another (and hopefully final) review.

chrismayer:
Thanks again for your review @KaiVolland! 

> Could you explain in one or two sentences (or maybe a source) why you decide to split the props interface in "default props" and "non default props"?

The "default props" are these props, which are optional and have a default value. The other have to be passed in by the father component. By modeling the "default props" in an own interface we can define a static ``defaultProps`` element in our component class to set the default values.
```
  public static defaultProps: DefaultAttributeComboProps = {
    label: 'Attribute',
    placeholder: 'Select Attribute',
  };
``` 

I found several sources describing this, e.g.:

https://stackoverflow.com/questions/37282159/default-property-value-in-react-component-using-typescript/37282264

Issue #69, opened 2018-05-17
UI for combined ComparisonFilter objects


chrismayer:
Since the GUI for a ``ComparisonFilter`` is ~~nearly~~ done (see #64) one of the next steps would be a GUI that allows the combination of several ``ComparisonFilter`` objects. Either with connected ``AND`` or ``OR``.

KaiVolland:
Solved with #498

PR #78, opened 2018-05-18
Implement a Rule GUI

chrismayer:
This introduces several React components to model a Rule-GUI.

The symbolizer GUI is not implemented yet and will be part of a follow-up PR (see issues #81 and #82).

A demo GUI showing the functionality is also part of this PR:

image

### DIFF
--- a/src/Component/Rule/NameField/NameField.tsx
+++ b/src/Component/Rule/NameField/NameField.tsx
@@ -29,7 +29,7 @@ class NameField extends React.Component<NameFieldProps, any> {
    * Extracts the text value of the ChangeEvent
    * and passes it to the passed in 'onChange' handler.
    */
-  valToString = (e: React.ChangeEvent<HTMLInputElement>) => {
+  onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const { value } = e.target;
     this.props.onChange(value);
   }
@@ -44,7 +44,7 @@ class NameField extends React.Component<NameFieldProps, any> {
           <Input
             className="gs-rule-namefield-input"
             value={this.props.value}
-            onChange={this.valToString}
+            onChange={this.onChange}
             placeholder={this.props.placeholder}
           />
 
--- a/src/Component/Rule/NameField/NameField.tsx
+++ b/src/Component/Rule/NameField/NameField.tsx
@@ -13,7 +13,7 @@ interface NameFieldProps extends Partial<DefaultNameFieldProps> {
 }
 
 /**
- * Input field for a textual filter value.
+ * Input field for the rule name.
  */
 class NameField extends React.Component<NameFieldProps, any> {
 
--- a/src/Component/Rule/NameField/NameField.tsx
+++ b/src/Component/Rule/NameField/NameField.tsx
@@ -2,6 +2,8 @@ import * as React from 'react';
 import { Input } from 'antd';
 import FormItem from 'antd/lib/form/FormItem';
 
+import './NameField.css';
+
 // default props
 interface DefaultNameFieldProps {
   label: string;
@@ -34,12 +36,12 @@ class NameField extends React.Component<NameFieldProps, any> {
   render() {
 
     return (
-      <div className="gs-rule-namefield" style={{ }} >
+      <div className="gs-rule-namefield" >
 
         <FormItem label={this.props.label} colon={false} >
 
           <Input
-            style={{ width: '100%' }}
+            className="gs-rule-namefield-input"
             onChange={this.valToString}
             placeholder={this.props.placeholder}
           />
--- a/src/Component/Rule/NameField/NameField.tsx
+++ b/src/Component/Rule/NameField/NameField.tsx
@@ -1,0 +1,54 @@
+import * as React from 'react';
+import { Input } from 'antd';
+import FormItem from 'antd/lib/form/FormItem';
+
+// default props
+interface DefaultNameFieldProps {
+  label: string;
+  placeholder: string;
+}
+// non default props
+interface NameFieldProps extends Partial<DefaultNameFieldProps> {
+  onChange: ((newValue: string) => void);
+}
+
+/**
+ * Input field for a textual filter value.
+ */
+class NameField extends React.Component<NameFieldProps, any> {
+
+  public static defaultProps: DefaultNameFieldProps = {
+    label: 'Name',
+    placeholder: 'Enter Rule Name'
+  };
+
+  /**
+   * Extracts the text value of the ChangeEvent
+   * and passes it to the passed in 'onChange' handler.
+   */
+  valToString = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+    this.props.onChange(value);
+  }
+
+  render() {
+
+    return (
+      <div className="gs-rule-namefield" style={{ }} >
+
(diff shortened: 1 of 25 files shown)


marcjansen:
Very nice. I'll let @KaiVolland give the final go, he is more involved.

One question, will we be able to combine n filters in one rule?

chrismayer:
Thanks for your review @marcjansen!
> One question, will we be able to combine n filters in one rule?

This is planned but not implemented yet. See #69.

src/Component/Rule/NameField/NameField.tsx
```diff
@@ -0,0 +1,54 @@
...
+}
+// non default props
+interface NameFieldProps extends Partial<DefaultNameFieldProps> {
+  onChange: ((newValue: string) => void);
+}
+
+/**
+ * Input field for the rule name.
+ */
+class NameField extends React.Component<NameFieldProps, any> {
```
KaiVolland: Have you thought about using `React.StatelessComponent<NameFieldProps>` for components without state?
I'm not sure about the benefits but as it exists it may have some. :smile:

src/Component/Rule/RemoveButton/RemoveButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+// non default props
+interface RemoveButtonProps extends Partial<DefaultRemoveButtonProps> {
+  ruleIdx: number;
+  onClick: ((ruleIdx: number) => void);
+}
+
+/**
+ * Button to remove a rule.
+ */
+class RemoveButton extends React.Component<RemoveButtonProps, any> {
```
KaiVolland: See comment above regarding stateless componenets.

src/Component/Rule/RemoveButton/RemoveButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  public static defaultProps: DefaultRemoveButtonProps = {
+    text: 'Remove Rule'
+  };
+
+  render() {
+
+    return (
+      <div className="gs-rule-removebutton" style={{ float: 'inherit' }} >
```
KaiVolland: Please don't use inline styles if you can avoid it.

src/Component/Rule/RemoveButton/RemoveButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        <Button
+          style={{}}
```
KaiVolland: Remove empty style object.

src/Component/Rule/Rule.tsx
```diff
@@ -0,0 +1,158 @@
+import * as React from 'react';
+
+import { Row, Col } from 'antd';
+import { ComparisonFilter, Rule as GsRule } from 'geostyler-style';
```
KaiVolland: You may want to add the `Gs` prefix to the `ComparisonFilter` too (or remove it from `GsRule`)...

As these are Interfaces we could think about using the more common `I` prefix for these to become: `IRule, IComparisonFilter` ...

src/Component/Rule/Rule.tsx
```diff
@@ -0,0 +1,158 @@
...
+import { Data as GsData } from 'geostyler-data';
```
KaiVolland: See comment above

src/Component/Rule/Rule.tsx
```diff
@@ -0,0 +1,158 @@
...
+      }
+    };
+
+    this.props.onRuleChange(rule, this.props.keyIndex);
+  }
+
+  render() {
+
+    return (
+      <div className="gs-rule" style={{ border: '1px solid lightgrey', padding: 10, margin: 10}} >
```
KaiVolland: Please move style to a stylefile (`Rule.css`).

src/Component/Rule/TitleField/TitleField.tsx
```diff
@@ -0,0 +1,54 @@
...
+   */
+  valToString = (e: React.ChangeEvent<HTMLInputElement>) => {
+    const { value } = e.target;
+    this.props.onChange(value);
+  }
+
+  render() {
+
+    return (
+      <div className="gs-rule-titlefield" style={{ }} >
```
KaiVolland: Remove empty style object.

src/Component/ScaleDenominator/MaxScaleDenominator.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  public static defaultProps: DefaultScaleDenominatorProps = {
+    label: 'Max. Scale',
+    placeholder: 'Enter max. Scale (Optional)'
+  };
+
+  render() {
+
+    return (
+      <div className="gs-max-scaledenominator" style={{ }} >
```
KaiVolland: Remove empty style object.

src/Component/ScaleDenominator/MinScaleDenominator.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  public static defaultProps: DefaultScaleDenominatorProps = {
+    label: 'Min. Scale',
+    placeholder: 'Enter min. Scale (Optional)'
+  };
+
+  render() {
+
+    return (
+      <div className="gs-min-scaledenominator" style={{ }} >
```
KaiVolland: Remove empty style object.

src/Component/Symbolizer/Preview/Preview.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * Symbolizer preview UI.
+ */
+class Preview extends React.Component<PreviewProps, any> {
+
+  render() {
+
+    return (
+      <div className="gs-symbolizer-preview" style={{ }} >
```
KaiVolland: Remove empty style object.

src/app/App.tsx
```diff
@@ -5,7 +5,8 @@ import 'antd/dist/antd.css';
 import './App.css';
 
 // import DataProviderUi from './Demo/DataProvider/DataProviderUi';
-import FilterDemo from './Demo/FilterUi/FilterDemo';
+// import FilterDemo from './Demo/FilterUi/FilterDemo';
```
KaiVolland: Outcommented lines can be removed.

src/Component/Rule/Rule.tsx
chrismayer: Since ``Rule`` would cause a naming conflict with the corresponding UI class I would prefer adding the ``Gs`` to the other import(s).

I would prefer the ``Gs``- prefix over ``I``-prefix because IMHO it points more to the actual content.

src/Component/Rule/NameField/NameField.tsx
chrismayer: Since this would cause bigger re-writes in the component definitions I suggest we postpone this until we are sure we have benefit from using ``StatelessComponent``.

chrismayer:
I addressed you review remarks. Passing in a Rule object caused some bigger changes since I had to rework the Rule-UI and its sub-components, so they can take an initial state. Would be cool, if you could have a look again, @KaiVolland.

chrismayer:
Thanks for reviewing again @KaiVolland. I pushed another commit , which renames the hadler fuctions as you suggested. I will merge now.

PR #84, opened 2018-05-22
Integrate OpenLayers incl. TS types

### LINKED ISSUE #83: Integrate OpenLayers with TypeScript
For several components, such as the style preview UI (see #81) and the preview map inside the ``StyleEditor`` (see #82) we need the integration of OpenLayers. Since we are using TypeScript we need some kind of TypeScript-binding for OpenLayers. Yet there is no official support for this but [here](https://github.com/openlayers/openlayers/issues/8120) can be found some ideas and infos.

chrismayer:
Integrates the OpenLayers lib incl. the TS types from https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ol

The ``@types/ol`` repo does not support the latest version of OL (v4.6.5). It only supports v4.6.1, but since OL uses semantic versioning I do not expect any problems by using OpenLayers v4.6.5.   

Fixes #83.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,6 +52,21 @@
       "integrity": "sha512-ekJ3mXJcJP+Nn5kC6eCmWPND/fHx/Ga12Lz0IJgTfGz1ge7OCIR5xcDY5tcYgnyM1kWsVDRH2bguxkGcNj39AQ==",
       "dev": true
     },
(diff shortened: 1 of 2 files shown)


chrismayer:
Thanks for reviewing @KaiVolland!

PR #91, opened 2018-05-23
Introduces styleguidist config

### LINKED ISSUE #59: Setup styleguidist
Setup styleguidist to provide the environment for examples and docs.

KaiVolland:
This introduces stylegudist. 

Closes #59

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,4 @@ node_js:
 - "8"
 script:
 - npm run test
-cache:
-  directories:
-  - node_modules
+cache: false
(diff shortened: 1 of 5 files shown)


KaiVolland:
Sections are more like chapters in the overview. It is not necessary but we should add a new section for every new Folder in the first order.

KaiVolland:
Test are all good on my local machine. Same node and npm version and i don't get why this fails on travis... I ll come back to this later.

KaiVolland:
Seems like the travis cache kept an old version of `babylon` which let the tests fail.
Compare: https://stackoverflow.com/a/41466909

PR #92, opened 2018-05-23
Add OL-map based symbolizer preview

chrismayer:
This adds a symbolizer preview realized as OpenLayers map. The preview component expects somes features (GeoJSON) and a symbolizer to define the style of the features on the map. 

Also a predefined OpenLayers map instance or an array of OpenLayers layers can be passed in to customized the setup of the map.

image

### DIFF
--- a/src/Component/Rule/Rule.tsx
+++ b/src/Component/Rule/Rule.tsx
@@ -5,6 +5,7 @@ import {
   Filter as GsFilter,
   ComparisonFilter as GsComparisonFilter,
   Rule as GsRule,
+  Symbolizer as GsSymbolizer
 } from 'geostyler-style';
 import { Data as GsData } from 'geostyler-data';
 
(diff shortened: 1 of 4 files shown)


chrismayer:
Thanks for reviewing @KaiVolland!

> As you can add a complete map, you could add one without the controls right?

Yes that's possible. Or would you prefer a map without (visible) controls as default?

KaiVolland:
Yes i would prefer the default to be just one feature with out controls and without a Backroundlayer. As I would like to add a preview of the complete style in a full map preview later on...

But this should not block this PR.

chrismayer:
I just added a further commit , which removes the controls and interactions from the default map of the
SymbolizerPreview. Therefore they can be configured by the new props ``controls`` and ``interactions``.

The background layer can be removed by setting the property ``showOsmBackground`` to ``false``.  

I am going to merge this now.

PR #105, opened 2018-05-25
Use GeoStyler OL parser in SymbolizerPreview

### LINKED ISSUE #96: Use GeoStyler OL parser in SymbolizerPreview
At the moment a temporary "proof-of-concept parser" to transform the internal styles to OpenLayers styles is used inside the ``SymbolizerPreview`` UI. Once we have an "official" ``geostyler-ol-parser`` we should replace the temporary parser with the official one. 

Therefore the transformation is encapsulated in the function ``symbolizer2OlStyle`` and can be replaced inside this function without any further API-breaks.

chrismayer:
This introduces the new [geostyler-openlayers-parser](https://www.npmjs.com/package/geostyler-openlayers-parser) so it can be used within the ``SymbolizerPreview`` in order to transform the symbolizer to an appropriate OpenLayers style. 

Fixes #96.


src/app/Demo/Rule/RuleDemo.tsx
```diff
@@ -139,6 +139,7 @@ class RuleDemo extends React.Component<FilterDemoProps, FilterDemoState> {
    */
   createStyle = () => {
     const style: GsStyle = {
+      name: 'GeoStylerStyle',
```
KaiVolland: This was the fix of #104 right? You could rebase this before merging...

chrismayer:
Thanks for your review @KaiVolland! I'll merge now.

PR #124, opened 2018-05-29
Introduces the (live) CodeEditor

KaiVolland:
This introduces the live CodeEditor.

It displays the current style representation and allows cross side editing of the style.
It includes a Format SelectField which allows the input of GeoStyler Style Parsers to edit the Style in the format you like (how awesome is that? :open_mouth:   ).

**GeoStyler Style:**
codeeditor_geostyler_style

**SLD:**
codeeditor_sld_style

According to testing the bidirectional synchronization some bugs regarding the manipulation of the immutable state object occurred. A fix is included that uses `cloneDeep` before setting new value on state properties.

This also includes a TemplateStyle that is loaded initially.

### DIFF
--- a/src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
+++ b/src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
@@ -38,7 +38,7 @@ interface ComparisonFilterState {
   attributeType?: string;
   operator: ComparisonOperator | undefined;
   value: string | number | boolean | null;
-  filter: ComparisonFilter | undefined;
+  filter: ComparisonFilter;
 }
 
 /**
@@ -57,6 +57,10 @@ class ComparisonFilterUi extends React.Component<ComparisonFilterProps, Comparis
       filter: nextProps.filter
     };
   }
+
+  public static defaultProps: DefaultComparisonFilterProps = {
+    filter: ['==', '', null]
+  };
 
   constructor(props: ComparisonFilterProps) {
     super(props);
@@ -105,7 +109,7 @@ class ComparisonFilterUi extends React.Component<ComparisonFilterProps, Comparis
         attribute: '',
         operator: undefined,
         value: null,
-        filter: undefined
+        filter: ComparisonFilterUi.defaultProps.filter
       };
     }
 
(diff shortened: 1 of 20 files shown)


src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
```diff
@@ -197,18 +197,11 @@ class ComparisonFilterUi extends React.Component<ComparisonFilterProps, Comparis
    * Stores the appropriate operator as member.
    */
   onOperatorChange = (newOperator: ComparisonOperator) => {
-
+    let filter: GsComparisonFilter = _cloneDeep(this.state.filter);
+    filter[0] = newOperator;
```
marcjansen: don't we need to clone? here and below?

src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
```diff
@@ -217,27 +210,17 @@ class ComparisonFilterUi extends React.Component<ComparisonFilterProps, Comparis
...
-    if (this.state.operator && this.state.attribute && newValue) {
-      const compFilter: ComparisonFilter = [
-        this.state.operator, this.state.attribute, newValue
-      ];
-      this.setState({filter: compFilter});
-
-      this.props.onFilterChange(compFilter);
-    }
+    let filter: GsComparisonFilter = _cloneDeep(this.state.filter);
+    filter[2] = newValue;
```
marcjansen: don't we need to clone?

marcjansen:
Awesome work here, everyone!

src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
KaiVolland: The `cloneDeep` is just need when manipulating the "immutable" state.

src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
KaiVolland: The `cloneDeep` is just need when manipulating the "immutable" state.

KaiVolland:
Thx for the reviews.
@marcjansen The polyfill comes from here: https://stackoverflow.com/questions/42213522/mocking-document-createrange-for-jest
I added the link to a comment in the code, too.

chrismayer:
This is very, very :cool: :exclamation:  :sunglasses:  :+1:

PR #142, opened 2018-06-06
Filter selectable attributes in AttributeCombo by a filter function

ahennr:
Title says it all.


ahennr:
Should fix https://github.com/terrestris/geostyler/issues/135

PR #180, opened 2018-06-12
Introduces Download button to CodeEditor

### LINKED ISSUE #178: Add a download / export button to GeoStyler application
It could be very useful to have a download / export button to get easy access to the style description. Especially for larger documents it could be nice simplification in usability.

KaiVolland:
This introduces a Button which allows to download the current value of the `CodeEditor` as a File.

It also removes the `geostyler-openlayers-parser` from the `StyleLoader` field of the App and adds a `label` prop to the Loaders.

Closes #178


KaiVolland:
Solved with https://github.com/terrestris/geostyler-geojson-parser/pull/17

package.json
```diff
@@ -25,6 +25,7 @@
     "antd": "3.6.2",
     "codemirror": "5.38.0",
     "color": "3.0.0",
+    "file-saver": "1.3.8",
```
chrismayer: Do we really need a lib for that? Not sure, just asking.

src/app/App.tsx
```diff
@@ -55,7 +55,6 @@ class App extends React.Component<AppProps, AppState> {
         <div className="settings">
           <StyleLoader
             parsers={[
-              OlStyleParser,
```
chrismayer: What is the exact reason, why you remove the OL Style Parser?

src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -164,18 +166,50 @@ class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState> {
...
+    const {
+      activeParser,
+      value
+    } = this.state;
+    const {
+      style
+    } = this.props;
+    if (style) {
+      let type = 'application/json;charset=utf-8';
+      if (activeParser && activeParser.title === 'SLD Style Parser') {
```
marcjansen: The comparison looks a bit fragile to my eyes...

src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -164,18 +166,50 @@ class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState> {
...
+        type = 'text/xml;charset=utf-8';
+      }
+      const blob = new Blob([value], {type});
```
marcjansen: Polyfill for `Blob` needed?

package.json
KaiVolland: The library covers multiple browsers. For modern Browsers we wouldn't need it.

src/app/App.tsx
KaiVolland: I removed it because there are actually no Ol Style Files that could be read. And as i didn't want the app to become to complex i decided to just remove the OlStyle as Upload Input.

src/Component/CodeEditor/CodeEditor.tsx
KaiVolland: Indeed. But as there is just a small amount of StyleParsers i think this is our best opportunity to determine which parser was choosen.
Furthermore the static `title` prop was introduced to identify the parser type.

PR #183, opened 2018-06-13
Fixes issues with BoolFilterField and NumberFilterField

### LINKED ISSUE #171: Activating BoolFilterField raises warning
When the ``BoolFilterField`` inside of a ComparisonFilter is activated the following warning appears: 

```
Warning: Failed prop type: Invalid prop `checked` supplied to `Checkbox`.
```
This has no functional impact on the application.

### LINKED ISSUE #173: Activating NumberFilterField crashes application
When the NumberFilterField inside of a ComparisonFilter is activated an exception is thrown and the application crashes:

image

Steps to reproduce:

1. Open https://terrestris.github.io/geostyler/
2. Load a GeoJSON with a numeric property
3. Select the numeric property in the ComparisonFilter UI

KaiVolland:
Fixes #173 - NumberFilterField
- Fixed import of `Form.Item`
- Removed min and max value as this should be moved to the validation and should not just block the numberspinner.

Fixes #171 - BoolFilterField
- Ensure that the passed checked prop is boolean

### DIFF
--- a/src/Component/Filter/BoolFilterField/BoolFilterField.tsx
+++ b/src/Component/Filter/BoolFilterField/BoolFilterField.tsx
@@ -62,7 +62,7 @@ class BoolFilterField extends React.Component<BoolFilterFieldProps, BoolFilterFi
         <Form.Item label={this.props.label} colon={false} >
 
           <Checkbox
-            checked={this.state.value}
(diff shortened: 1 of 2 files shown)


KaiVolland:
I added a corresponding issue: #188

PR #184, opened 2018-06-13
Introduces onCheckChange listeners for FieldSets

### LINKED ISSUE #175: Consequence of deactivation of scale section
In case the ``Use Scale`` fieldset is deactivated the ``CodeEditor`` does not change. One could expect that the deactivation removes the scale-section in the filter code. If this is the wanted behavior the synchronization between UI and the code editor has to be done. If not we should think about changing the label.

### LINKED ISSUE #176: Consequence of deactivation of filter section
In case the ``Use Filter`` fieldset is deactivated the ``CodeEditor`` does not change. One could expect that the deactivation removes the scale-section in the filter code. If this is the wanted behavior the synchronization between UI and the code editor has to be done. If not we should think about changing the label.

KaiVolland:
This introduces the `onCheckChange` prop for the `FieldSet` and adds corresponding functions for the `Filter` and `ScaleDenominator` `FieldSet`s.

Fixes #175 
Fixes #176

### DIFF
--- a/src/Component/Rule/Rule.tsx
+++ b/src/Component/Rule/Rule.tsx
@@ -9,7 +9,9 @@ import {
   // Filter as GsFilter,
   ComparisonFilter as GsComparisonFilter,
   Rule as GsRule,
-  Symbolizer as GsSymbolizer
+  Symbolizer as GsSymbolizer,
+  Filter as GsFilter,
+  ScaleDenominator as GsScaleDenominator
 } from 'geostyler-style';
 import { Data as GsData } from 'geostyler-data';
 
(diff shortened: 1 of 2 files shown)


src/Component/Rule/Rule.tsx
```diff
@@ -156,13 +198,19 @@ class Rule extends React.Component<RuleProps, RuleState> {
             />
           </div>
           <div className="gs-rule-right-fields" >
-            <Fieldset title="Use Scale">
+            <Fieldset
+              title="Use Scale"
```
marcjansen: Should this be a prop?

src/Component/Rule/Rule.tsx
```diff
@@ -156,13 +198,19 @@ class Rule extends React.Component<RuleProps, RuleState> {
...
+              onCheckChange={this.onScaleCheckChange}
+            >
               <ScaleDenominator
                 scaleDenominator={rule.scaleDenominator}
                 onChange={this.onScaleDenominatorChange}
               />
             </Fieldset>
-            <Fieldset title="Use Filter">
+            <Fieldset
+              title="Use Filter"
```
marcjansen: Should this be a prop?

PR #185, opened 2018-06-13
Add a simple editor for text symbolizer

chrismayer:
This adds a simple editor for ``TextSymbolizer``s, which allows to edit the

  - text color
  - font type
  - text opacity
  - text size
  - text offset

of the label / text:

image


KaiVolland:
I'll merge so we can make a new release.

PR #193, opened 2018-06-13
Add select field for label attribute in TextEditor

chrismayer:
This adds an `AttributeCombo` component to the `TextEditor` in order to let the user select the field / attribute which is used as label in the map.

Herewith the `SymbolizerPreview` is changed that it accepts the full internal data object (instead of just the features), so the schema can be forwarded to the new `AttributeCombo`.

At the moment the preview only works for string based fields. A separate issue will opened since this could be a problem in the [geostyler-openlayers-parser](https://github.com/terrestris/geostyler-openlayers-parser) .

image

### DIFF
--- a/src/Component/Symbolizer/TextEditor/TextEditor.tsx
+++ b/src/Component/Symbolizer/TextEditor/TextEditor.tsx
@@ -21,6 +21,7 @@ import './TextEditor.css';
 
 // default props
 interface DefaultTextEditorProps {
+  fieldLabel: string;
   opacityLabel: string;
   colorLabel: string;
   sizeLabel: string;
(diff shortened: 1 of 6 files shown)


chrismayer:
The mentioned problem with non-string fields is now documented here https://github.com/terrestris/geostyler/issues/194.

src/Component/Symbolizer/TextEditor/TextEditor.tsx
```diff
@@ -72,7 +78,19 @@ class TextEditor extends React.Component<TextEditorProps, {}> {
     }
 
     return (
+
       <div className="gs-text-symbolizer-editor" >
+         <div className="editor-field width-field">
```
KaiVolland: Copy&Paste in ClassName

chrismayer:
> I'd prefer the dummy text Foo Fighters next time 

Absolutely! :metal: 

Thanks for the reviews @KaiVolland and @marcjansen.

PR #216, opened 2018-06-15
Align DataInput UIs when upload button is shown

chrismayer:
This prevents the shift of the select field due to the shown upload button. So the fields keep aligned, even one DataInput shows the upload button and one does not.

Before:

image

After:

image


KaiVolland:
Thanks. Please merge.

PR #221, opened 2018-06-15
Use longer default LineString in SymbolizerPreview

chrismayer:
In order to have a better visual representation for `LineSymbolizers` this adds a longer default LineString (especially for visualizing line-dashes).

image


chrismayer:
Thanks for reviewing, bro!

PR #222, opened 2018-06-15
Introduces the wfs-data-parser as an option for the DataLoader

### LINKED ISSUE #213: Add WFS-Parser to GeoStyler app
Since we have a [WFS Data Parser](https://github.com/terrestris/geostyler-wfs-parser) for GeoStyler we could think about integrating this into the GeoStyler app. This would be another great (real world) use case. 

Any opinions on this? Vetos?

KaiVolland:
This introduces the `geostyler-wfs-parser` as option for the `DataLoader`.

localhost_3000_ 10

This form is shown in modal when selecting the `WFS Data Parser` option.

Closes #213


KaiVolland:
This profits from https://github.com/terrestris/geostyler-wfs-parser/pull/61

src/Component/DataInput/WfsParserInput/WfsParserInput.css
```diff
@@ -0,0 +1,19 @@
...
+  flex: 2;
+}
+
+.ant-row .ant-form-item-control-wrapper {
+  flex: 3;
+}
+
+.ant-modal-footer {
+  display: none;
+}
```
dnlkoch: Is this global override intended?

src/Component/DataInput/WfsParserInput/WfsParserInput.tsx
```diff
@@ -0,0 +1,269 @@
...
+  Button,
+  Input,
+  Select,
+  InputNumber
+} from 'antd';
+const Option = Select.Option;
+
+import {
+  get as _get
+} from 'lodash';
```
dnlkoch: For build size optimization try to import `get` from `lodash/get`.

PR #246, opened 2018-06-25
Replace name as key for Rule UI

### LINKED ISSUE #234: Enhance typing in rules name field
When one types in the name field of the rule of the [example application](https://terrestris.github.io/geostyler/), the textfield looses the focus, and the map gets reloaded completely.

It'd be nice 
* if the field didn't loose focus, so one can type more than one character
* if the map would not do a full reload

chrismayer:
This sets the key for the Rule UIs to somethings which is not related to the rule's name. Otherwise the Rule UI is re-created every time the user changes the rule name.

Fixes #234.


marcjansen:
Thanks for this

PR #268, opened 2018-07-03
Fixes tests on CI

KaiVolland:
This adds [`--runInBand`](http://jestjs.io/docs/en/cli#runinband) to the test script to avoid tests timing out on Travis CI.


marcjansen:
Find ich gut

PR #277, opened 2018-07-13
Pass dataProjection to down to preview

ahennr:
This PR a further prop `dataProjection` to `Style` and `Rule` that is passed to`Preview` component.

Please review.


chrismayer:
LGTM, thanks!

PR #294, opened 2018-07-25
Add internationalisation

jansule:
Internationalisation now works for built-in antd components as well as dedicated geostyler components. A first draft of German and English translations was created (please take a look at the german translations, especially) . Two buttons were added to the view allowing language switching. By default, English will be displayed. The buttons were only created as a proof that everything works. Thus, the styling is not final at all.

To communicate the current locale throughout all components, context `antLocale` from antd is being used. A custom HOC was created to pass the current locale to the components, respectively. German and English language packages, as well as corresponding description files, were added to `src/locale/`. 

So far, following steps need to be done to extend/use language packages for new components:
1. Add variable name descriptions to `src/locale/<language>.d.ts` 
  **Note:** To avoid naming conflicts, at root level, extension must have following structure: `"Gs" + <ComponentName>`. E.g. GsApp
2. Add key value pairs to  `src/locale/<language>.js`
3. In `<component>.tsx` add `import { localize } from '<path>/<to>/Component/LocaleWrapper/LocaleWrapper`
4.  Create interface `<componentName>Locale` and add description for used keys
5. Add prop `locale?: <componentName>Locale` to default props of component
6. <del>Create `constructor(props: any)` in component</del>
7. Add `{ locale } = this.props` in `render()` in component, to use translations
8. Export component with HOC. E.g. `export default localize(<ComponentName>);`

<del>**Important:** In order to receive the context, `compilerOptions.module` in `tsconfig.json` was set from `commonjs` to `esnext`. This might lead to unexpected behaviour in the future.</del>

Edit: Remove Step 6 due to [this Commit](https://github.com/terrestris/geostyler/pull/294/commits/82526bd0135dada7f503e3602a0befb2203e328e).


jansule:
There were some problems with defining the signature for the HOC in `Components/LocaleWrapper` and the injection of the `locale` prop.  Right now it works, but `locale` needed to be set to optional to make everything work. (Otherwise, either a constructor with `props: any` needed to be added to every component using that HOC, or the `locale` property needed to be set already in the `render()` method.) 

This might be a temporary solution, but in my opinion it does not make sense to set `locale` to optional, since injecting this property is the only thing the HOC does.

@chrismayer could you take a look at this? @dnlkoch, @KaiVolland and I could not find a solution yet.

KaiVolland:
A screenshot would be nice enhancement to the PR.

src/index.tsx
```diff
@@ -1,9 +1,75 @@
...
+  GsCodeEditor: object;
+  GsPreview: object;
+  GsCircleEditor: object;
+  GsColorField: object;
+  GsFillEditor: object;
+  GsIconEditor: object;
+  GsLineEditor: object;
+  GsTextEditor: object;
+  GsKindField: object;
+  GsScaleDenominator: object;
```
KaiVolland: You might use the Locale Props here instead of `object`.
```typescript
GsRule: RuleLocale
…
```

jansule:
English
grafik

German
grafik

src/Component/Rule/Rule.tsx
```diff
@@ -18,30 +18,46 @@ import { Data as GsData } from 'geostyler-data';
...
+
 import {
   cloneDeep as _cloneDeep
 } from 'lodash';
 
+// i18n
+export interface RuleLocale {
+    removeRuleBtnText: string;
+    scaleFieldSetTitle: string;
+    filterFieldSetTitle: string;
```
chrismayer: I think the `Set` is not necessary in `scaleFieldSetTitle` and `filterFieldSetTitle`

src/locale/de_DE.js
```diff
@@ -0,0 +1,102 @@
...
+            Circle: 'Kreis',
+            Fill: 'Füllung',
+            Icon: 'Icon',
+            Line: 'Linie',
+            Text: 'Text'
+        }
+    },
+    GsScaleDenominator: {
+        minScaleDenominatorLabelText: 'Min. Maßstab',
+        maxScaleDenominatorLabelText: 'Max. Maßstab',
```
chrismayer: To be very tiny this must be something like `Min. Maßstabszahl`

chrismayer: To be very tiny this must be something like `Max. Maßstabszahl`

chrismayer:
Great! Very good job @jansule :+1: 
I had only some minor (non-blocking) comments, but I agree with @KaiVolland that we should be careful with changing to `esnext`. We first had `module` option set to another value and changed it by purpose to `commonjs` for several reasons I do not remember anymore. I had a quick look but I did not find the corresponding PR / commit. If we are sure this has no side effects I am +1 for merging this.

jansule:
@chrismayer made the changes you proposed. Sorry for the messed up commit history. I'm not very familiar with rebasing, but should not happen again in the future.

jansule:
Could reset `tsconfig.json` without running into the same problems as before. Thus, `compilerOptions.module` now is set to `commonjs` again and hence, should not create any side effects.

jansule:
@dnlkoch merged conflicts.

PR #300, opened 2018-07-30
Optimize imports of lodash and openlayers

weskamm:
Replaces the `openlayers` module with the `ol` module.
Also fixes some lodash imports to avoid exploding file sizes of production builds

@jansule please review


weskamm:
thanks, merging now

PR #338, opened 2018-08-13
Different tslint for Development and Production

jansule:
Now two tslint configs exist. One for development, one for production. In development, the defaultSeverity is set to `warning`. Thus, compiling does not fail when breaking a rule. In production defaultSeverity is set to `error`, resulting in failing compilation.

When running `npm run build` tests run before building.

Solution for config setup taken from
https://github.com/wmonk/create-react-app-typescript/issues/329#issuecomment-411329453
https://github.com/wmonk/create-react-app-typescript/issues/133#issuecomment-393635056


package.json
```diff
@@ -49,8 +49,9 @@
     "build:styleguide": "styleguidist build",
     "eject": "react-scripts-ts eject",
     "lint": "tslint --project tsconfig.json --config tslint.json && tsc --noEmit --project tsconfig.json",
-    "pretest": "npm run lint",
-    "prebuild": "rimraf dist/** && rimraf build/**",
+    "lint:prod": "tslint --project tsconfig.json --config tslint.prod.json && tsc --noEmit --project tsconfig.json",
+    "pretest": "npm run lint:prod",
+    "prebuild": "rimraf dist/** && rimraf build/** && npm run test",
```
dnlkoch: In the current CI configuration `npm run test` will be called before `npm run build` as well. Please remove it either here or in `.travis.yml` otherwise it will be called twice in CI.

PR #349, opened 2018-08-15
Mapping function for titles in operator combo and disable titles

ahennr:
This PR adds following props to `OperatorCombo`
* show title of selected item (when selected item hovered) -> `boolean`
* mapping function for operator titles


ahennr:
Thx for review

Merging.

PR #350, opened 2018-08-15
Adds attribute name mapping function

ahennr:
Title says it all.


ahennr:
thx for review.

Merging.

PR #383, opened 2018-08-24
Add unit tests for Style Component

jansule:
add helper function to TestUtil.tsx
export class Style to be able to run shallow tests without localeWrapper


KaiVolland:
Nice one. Please rename `TestUtil.tsx` to `TestUtil.ts` as `tsx` is only for files that export react components.

KaiVolland:
ping @jansule

jansule:
@KaiVolland I did not rename the tsx because the linter does not like the html tags within the shallow functions, when renaming to `.ts`. Even though, they are properly written.

PR #411, opened 2018-09-10
Add `checked` prop to FieldSet

jansule:
Removed the `visible` state in FieldSet and added prop `checked` instead. This property accepts a boolean and depending on the condition renders the content of FieldSet.

### DIFF
--- a/src/Component/Rule/Rule.tsx
+++ b/src/Component/Rule/Rule.tsx
@@ -78,6 +78,8 @@ interface RuleState {
   symbolizerEditorVisible: boolean;
   storedFilter?: GsFilter;
   storedScaleDenominator?: GsScaleDenominator;
+  scaleFieldChecked?: boolean;
+  filterFieldChecked?: boolean;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


src/Component/Rule/Rule.tsx
```diff
@@ -78,6 +78,8 @@ interface RuleState {
   symbolizerEditorVisible: boolean;
   storedFilter?: GsFilter;
   storedScaleDenominator?: GsScaleDenominator;
+  scaleFieldChecked?: boolean;
+  filterFieldChecked?: boolean;
```
KaiVolland: It may make sense to add this booleans to the `getDerivedStateFromProps` and if the rule has no `ScaleDenominator`/`Filter` set it to `false` otherwise to `true`. As `getDerivedStateFromProps` is called on every render you might need to test around …

jansule: @KaiVolland made requested changes. Since ScaleDenominator and Filter are not being removed completely from a geostyer-style when being empty, the behaviour is as expected. The Fieldsets will automatically uncheck only if you remove the property from within the code editor.

PR #452, opened 2018-10-01
Add config for building browser library with webpack 3

### LINKED ISSUE #419: GeoStyler as browser library
To use the `GeoStyler` components outside of an react application (from the npm universe) we need to do a seperate build.
As we did create the project with `create-react-app-typescript` we are stuck to WebPack v3. 

#413 does export the GeoStyler as a browser compatible library but breaks the regular build (WebPack v4).

So a solution is needed to bundle the tsx and css files with the current setup.
Alternatively we could 'eject' `GeoStyler`...

Any ideas @terrestris/devs ?

jansule:
This PR enables building of a browser library that allows using geostyler outside of the react universe. An existing PR for webpack 4 already exists [here](https://github.com/terrestris/geostyler/pull/413) as well as a corresponding [issue](https://github.com/terrestris/geostyler/issues/419), which will be partially solved through this PR. 

`browser-build.config.js` was created which is a simple webpack config. It was renamed so it is clear that it will only be used for creating the browser version. According script was added to package.json and can be called via `npm run build:browser`. This was also integrated into the `npm run build` script.

Since we are still using webpack 3, `create-react-app-typescript` still runs without errors.

The build output is available in `/browser`.

Also exported a few missing components.

Closes #419

### DIFF
--- a/browser-build.config.js
+++ b/browser-build.config.js
@@ -1,4 +1,5 @@
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
+const webpack = require("webpack");
 module.exports = {
   entry: "./src/index.js",
   output: {
@@ -43,7 +44,8 @@ module.exports = {
     ]
   },
   plugins: [
-    new ExtractTextPlugin("geostyler.css")
+    new ExtractTextPlugin("geostyler.css"),
+    new webpack.optimize.UglifyJsPlugin(),
   ],
   // When importing a module whose path matches one of the following, just
   // assume a corresponding global variable exists and use that instead.
--- a/browser-build.config.js
+++ b/browser-build.config.js
@@ -1,51 +1,56 @@
 const ExtractTextPlugin = require("extract-text-webpack-plugin");
- module.exports = {
+module.exports = {
   entry: "./src/index.js",
   output: {
-      filename: "geostyler.js",
-      path: __dirname + "/browser",
-      library: 'GeoStyler'
+    filename: "geostyler.js",
+    path: __dirname + "/browser",
+    library: "GeoStyler"
   },
-   // Enable sourcemaps for debugging webpack's output.
-  devtool: "source-map",
-   resolve: {
-      // Add '.ts' and '.tsx' as resolvable extensions.
-      extensions: [".ts", ".tsx", ".js", ".json"]
+  resolve: {
+    // Add '.ts' and '.tsx' as resolvable extensions.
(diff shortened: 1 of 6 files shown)


browser-build.config.js
```diff
@@ -0,0 +1,51 @@
...
+  ],
+   // When importing a module whose path matches one of the following, just
+  // assume a corresponding global variable exists and use that instead.
+  // This is important because it allows us to avoid bundling all of our
+  // dependencies, which allows browsers to cache those libraries between builds.
+  externals: {
+      "react": "React",
+      "react-dom": "ReactDOM"
+  }
+};
```
dnlkoch: Please homogenize indentation and use of single/double quotes.

browser-build.config.js
```diff
@@ -0,0 +1,51 @@
+const ExtractTextPlugin = require("extract-text-webpack-plugin");
+ module.exports = {
+  entry: "./src/index.js",
+  output: {
+      filename: "geostyler.js",
+      path: __dirname + "/browser",
+      library: 'GeoStyler'
+  },
+   // Enable sourcemaps for debugging webpack's output.
+  devtool: "source-map",
```
dnlkoch: I'm not sure if we really want source maps in production as they would bloat the output/build size.

browser-build.config.js
```diff
@@ -0,0 +1,51 @@
...
+            loader: 'babel-loader',
+            query: {
+              presets: ['@babel/env']
+            }
+          }
+      ]
+  },
+   plugins: [
+    new ExtractTextPlugin("geostyler.css")
+  ],
```
dnlkoch: Is there also a default optimization config in older webpack versions (<4) present? E.g. for `UglifyJS`?

browser-build.config.js
```diff
@@ -0,0 +1,51 @@
+const ExtractTextPlugin = require("extract-text-webpack-plugin");
+ module.exports = {
+  entry: "./src/index.js",
+  output: {
+      filename: "geostyler.js",
+      path: __dirname + "/browser",
```
dnlkoch: Please add the `browser` directory to the [files entry](https://github.com/terrestris/geostyler/blob/master/package.json#L6) in the `package.json`, otherwise it wouldn't be copied to npm.

dnlkoch: And also add the path to the `.gitignore`.

jansule:
@dnlkoch could you review new changes?

PR #483, opened 2018-10-16
fix(package): update antd to version 3.10.1

KaiVolland:
Closes #453


jansule:
antd 3.10.1 is using typescript3, whereas we are using version 2.9.x. MediaQueryListEvent was introduced in typescript3. Hence, checks fail. See here https://github.com/ant-design/ant-design/issues/12655. Btw, antd depends on typescript3 since v3.8.x.

PR #498, opened 2018-10-22
Introduces FilterTree for complex filters

KaiVolland:
This introduces the `FilterTree` Component.

It adds support for editing complex/interleaved filters.

filter_tree

TODO:

~~- [ ] Refactor `TreeNode`s to new component~~ Canceld as there is to much shared code
- [x] Fix drag & drop issues

### DIFF
--- a/src/Component/Filter/FilterTree/FilterTree.tsx
+++ b/src/Component/Filter/FilterTree/FilterTree.tsx
@@ -120,7 +120,7 @@ class FilterTree extends React.Component<FilterTreeProps, FilterTreeState> {
     );
 
     const addButton = (
-      <Tooltip title="Add Filter" placement="left">
+      <Tooltip title="Add Filter" placement="top">
         <Dropdown
           trigger={['click']}
           overlay={addFilterMenu}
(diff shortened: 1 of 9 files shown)


src/Component/Filter/FilterTree/FilterTree.tsx
```diff
@@ -0,0 +1,567 @@
...
+} from 'geostyler-style';
+
+import './FilterTree.css';
+
+import {
+  Data as Data
+} from 'geostyler-data';
+
+import ComparisonFilterUi from '../ComparisonFilter/ComparisonFilter';
+import { TreeNode } from 'antd/lib/tree-select';
```
jansule: You already defined TreeNode in line 15. Looks like you are running into naming conflicts.

src/Component/Filter/FilterTree/FilterTree.tsx
```diff
@@ -0,0 +1,567 @@
...
+          overlay={changeFilterMenu}
+        >
+          <Button size="small">
+            <Icon type="filter"/>
+          </Button>
+        </Dropdown>
+      </Tooltip>
+    );
+
+    if (operator === '&&') {
```
jansule: Minor note: A switch statement could improve the readability and maintainability, here.

src/Component/Filter/FilterTree/FilterTree.tsx
```diff
@@ -0,0 +1,567 @@
...
+  /**
+   * Transforms a position String like '[2][3]' to an positionArray like [2, 3].
+   */
+  positionStringAsArray = (positionString: string) => {
+    return positionString
+      .replace('][', ',')
+      .replace(']', '')
+      .replace('[', '')
+      .split(',')
+      .map(i => parseInt(i, 10));
```
jansule: This will break if using more than 9 filters, because the map function will parse over each digit instead of each number.
`'[10][5]' => [1,0,5]`

src/Component/Filter/FilterTree/FilterTree.tsx
```diff
@@ -0,0 +1,567 @@
...
+  }
+
+  /**
+   * Transforms am positionArray like [2, 3] to a string like '[2][3]'.
+   */
+  positionArrayAsString = (positionArray: number[]) => {
```
jansule:
```suggestion
  positionArrayAsString = (positionArray: [number, number]) => {
```
Through this change, positionArray is a tuple of two numbers instead of an arbitrary array. This will make sure, you only have an array of two numbers. Any other structure will fail. Also useful, because .replace() only replaces the first occurence of the string if not used with a regexp with global flag.

KaiVolland: I fixed the replace method in 6b1f652

src/Component/Filter/FilterTree/FilterTree.tsx
KaiVolland: Line 387 replaces `][` with `,`
Line 388 removes `[`
Line 389 removes `]`
Line 390 splits the string by the added `,` from Line 387

So `'[10][3]'` becomes `["10", "3"]` to finally become `[10, 3]`.

PR #508, opened 2018-10-25
WIP: handle scaleDenominator in preview

jansule:
**Important:** Please wait with merging this PR until
- [this PR of geostyler-openlayers-parser]()
was released.

Preview now applies defined scaleDenominators to map. When changing a style, map will always center on the feature and zoom to a resolution within defined scaleDenominator. If no scaleDenominator was defined, map will zoom to fit extend of feature. Also added zoom and pan interaction to map in demo. So a user can check the ranges where style will be applied and where not, depending on scaleDenominator.

Also made small changes to eslint files, to allow unused variables in dev mode.

half-opt

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,1 @@
+declare module '@terrestris/ol-util/dist/MapUtil/MapUtil';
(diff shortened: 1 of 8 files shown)


jansule:
@marcjansen I guess you mean instead of the automated rescaling?

jansule:
This PR will never be merged due to current refactoring of `Preview` and `Renderer` components. However, the new features in this PR could be used for future inspiration, when refactoring is finished. -> PR stays open until then.

KaiVolland:
I'll close this PR. Code doesn't get lost though...

Issue #504, opened 2018-10-25
Symbolizer Fields loose focus


KaiVolland:
The symbolizer form fields loose there focus which makes it impossible to enter values.

jansule:
Fixed in #509.

PR #509, opened 2018-10-25
Fix EditorWindow Drag Bug

jansule:
The drag event on the whole window lead to loosing focus of input fields (related issue here #504). Problem was fixed through limiting the drag event to the EditorWindow Titlebar. Also Cursor style is set to pointer when hovering over Titlebar.

dragwindow


marcjansen:
Why is the cursor not set to grabbing or to movable? would be even better, right?

jansule:
Good point. Will change this.

marcjansen:
Nice, thx. Only the gif is now outdated :smile:

jansule:
Done

PR #510, opened 2018-10-26
chore(package): update react-styleguidist to version 8.0.2

dnlkoch:
Follow up of #500.


jansule:
As long as we are using webpack 3, the checks will fail. styleguidist depends on webpack 4.

PR #527, opened 2018-10-31
Introduces SLDRenderer

### LINKED ISSUE #503: SLD Preview
A preview component to show the real results of an SLD should be added.
If the target of the GeoStyler is an SLD the current Preview (OpenLayers) differs to much from the of an SLD.

TODOs:
- Create a configurable Componenet `SLDPreview`
   Props should be `serverUrl`, `featureType`, `style`, …
- Introduce a way to choose which preview should be used.

KaiVolland:
This introduces the `SLDRenderer` component (currently not included in the demo Application). It does a getLegendRequest with a mocked SLD.

It requires a `wmsBaseUrl` and a `layer` as prop.
Eventhough we don't need a layer for our purpose it is required by the standard.

Now advanced styles can be shown, too. :tada: 

sldpreview

**Only tested with GeoServer**
closes #503


jansule:
Could you add a delay to the SLDRenderer so that a request will not be fired on every single change?

src/Component/Symbolizer/SLDRenderer/SLDRenderer.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    if (!_isEqual(symbolizers, prevProps.symbolizers)) {
+      this.setLegendGraphicUrlForRule(symbolizers);
+    }
+  }
+
+  /**
+   * The function that sets the legends graphic for each rule
+   *
+   * @param {Object} rule The Sld rule
```
ahennr: JSDoc params doesn't match params of method.

PR #546, opened 2018-11-06
[WIP] Compact view

KaiVolland:


src/Component/RuleTable/RuleTable.tsx
```diff
@@ -0,0 +1,297 @@
...
+
+  minScaleRenderer = (text: string, record: RuleRecord, index: number) => {
+    const minScaleDenominator = _get(record, 'scaleDenominator.min');
+    const value = minScaleDenominator ? parseFloat(minScaleDenominator) : undefined;
+    return (
+      <InputNumber
+        value={value}
+        min={0}
+        formatter={val => `1:${val}`}
+        parser={(val: string) => parseFloat(val.replace('1:', ''))}
```
jansule: You could use the property `addonBefore` instead `formatter` and `parser` here. This adds a label before the input, like this:

image

Then you would not have to format and parse the scales and empty inputs would still look fine.

jansule:
Will there also be some tests in the future?

src/Component/RuleTable/RuleTable.tsx
KaiVolland: Yeah i saw that. Still not sure what looks better...

KaiVolland:
I will have a look at tests and coverage tomorrow in a followup.

KaiVolland:
@jansule small update with https://github.com/terrestris/geostyler/pull/546/commits/609c71f0eb57b9fdde200e90bf8921ef75afa379 could you review that too? I'll merge afterwards.

PR #557, opened 2018-11-07
Tests and code-care

KaiVolland:
- Adds missing tests for every `Component` and `Util`
- Makes changeListeners optional
- Exports props of every component
- Add named export for every component and use these in Tests
- Fix launch.json to be able to debug tests in VSCode


src/Component/ScaleDenominator/ScaleDenominator.tsx
```diff
@@ -21,12 +22,16 @@ export interface ScaleDenominatorLocale {
   maxScaleDenominatorPlaceholderText?: string;
 }
 
+interface ScaleDenominatorDefaultProps {
+  locale?: ScaleDenominatorLocale;
+}
+
 // non default props
-interface ScaleDenominatorProps {
+export interface ScaleDenominatorProps extends Partial<ScaleDenominatorDefaultProps> {
```
jansule: We should use a consistent name for our default props. Sometimes it's `DefaultCOMPONENTNAMEProps` sometimes it's `COMPONENTNAMEDefaultProps`. It would be nice if this could be applied in this PR as well.

src/Component/Symbolizer/Preview/Preview.spec.tsx
```diff
@@ -13,11 +13,12 @@ describe('Preview', () => {
   }];
 
   beforeEach(() => {
-    wrapper = TestUtil.shallowRenderComponent(Preview, {
+    const props: PreviewProps = {
       locale: en_US.GsPreview,
-      onSymbolizerChange: jest.fn(),
+      onSymbolizersChange: jest.fn(),
```
jansule: Why not declaring `let onSymbolizersChangeDummy: jest.Mock;` and assigning it to `onSymbolizersChange`? Would be consistent with other tests and easier to spy on, if tests will be extended.

src/Component/Symbolizer/Renderer/Renderer.spec.tsx
```diff
@@ -12,10 +12,11 @@ describe('Renderer', () => {
   }];
 
   beforeEach(() => {
-    wrapper = TestUtil.shallowRenderComponent(Renderer, {
-      onSymbolizerChange: jest.fn(),
+    const props: RendererProps = {
+      onClick: jest.fn(),
```
jansule: see comment above

PR #560, opened 2018-11-08
Update wfs-parser and geostyler-data

jansule:
Temporarily using `beta` releases of geostyler-wfs-parser and geostyler-data that fixes CI fails. Need to update manually, since greenkeeper only checks for tag `latest`.


KaiVolland:
Do we need to merge this, or is this just for testing purpose?

jansule:
Sorry, too fast =)
It makes sense to merge it. So all geostyler-related packages have clean dependency setups. As soon as other dependency updates or features will be added to related packages, we can make a proper release.

PR #572, opened 2018-11-12
[WIP] Adds colorpicker to CodeEditor

KaiVolland:


KaiVolland:
Seems like the "codemirror-colorpicke" is still somewhat buggy. We should may wait for the 2.0 release...

Issue #526, opened 2018-10-31
To eject or not eject – That is the question


KaiVolland:
We're more and more getting problems due to the CRA (Create-React-App) setup.

1. Currently there is no (clean) way to include a CSS preprocessor like LESS. #65 
2. Updating devdependencies like webpack, jest, react-styleguidist relies to CRA. This is can be sometimes blocking.
3. It feels like a blackbox.

jansule:
Just as an example: Currently, we cannot update webpack to version 4 because react-scripts-ts depends on webpack 3 (#489). This also blocks update of styleguidist (#510).

chrismayer:
I am not too deep into the CRA topic but the points listed above by @KaiVolland give me the feeling that it would be better to eject. Maybe the eject with a proper and complete documentation won't be that problematic?

KaiVolland:
Ejecting is simple and just leaves us some config files (webpack, babel, etc.) that can be shrinked afterwards by removing some uneeded configs. Shouldn't be a problem at all and IMHO meanwhile the benefits of the flexibility might outwheigh the comfort of CRA.

KaiVolland:
I'll giv it a try.

KaiVolland:
I ejected an ran into some problems when building.
I pushed the current state: https://github.com/KaiVolland/geostyler/tree/eject but as the priority is not THAT high... I'll skip this for know.

![https://thumbs.gfycat.com/NecessaryCourteousFalcon-size_restricted.gif](https://thumbs.gfycat.com/NecessaryCourteousFalcon-size_restricted.gif)

jansule:
Currently, I am moving the demo application into a dedicated repo [geostyler-demo](https://github.com/terrestris/geostyler-demo) to make a clear distinction between the library and the demo application. Through this, we can completely remove CRA from this repo and handle it the same way as we handle react-geo. Thus, we do not have to think about whether to eject or not.

The problems with CRA will then not affect geostyler anymore.

KaiVolland:
I think this process is finished. Reopen if im wrong.

Issue #65, opened 2018-05-17
Enable less or similar for advanced styling


chrismayer:
We should integrate a CSS Preprocessor such as less to get advanced styling possibilities.

See: https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-css-preprocessor-sass-less-etc

KaiVolland:
I don't like the approach recommende by create-react-app as linked above. This would lead to build files (css) inside the module-folders.
We could think about ejecting (and getting rid of CRA completly).

jansule:
As soon as we got rid of CRA, adding a CSS preprocessor should not be a problem. Little more info in #526

PR #574, opened 2018-11-12
Remove demo & CRA

### LINKED ISSUE #459: Externalize Demo and add peerDependencies
I think it would be beneficial for the understanding of geostyler to externalise the demo into its own repository and only keep the actual components in this repo. Trough this, it might become clearer what geostyler actually is - a library.

Another advantage is that we could add react, openlayers and antd as peerDependencies. This could resolve issues when using geostyler within an environment that depends on other (minor) versions of those packages. Especially, because we are using only fix versions of those packages.

@terrestris/devs what do you think about this?

jansule:
Demo will be moved to a dedicated repository, leaving this as a pure library. This means, CRA is not needed anymore. 

- Removed related packages from dependencies and added required ones to make everything run without CRA. 
- Removed unneded files (e.g. App.tsx)
- Overall structure did not change (yet?)
- Added ol, react, antd as peerDependencies

Now everything runs on webpack 4. Already updated styleguidist to latest version.

**If merging this PR, there will be no Demo available until terrestris/geostyler-demo was set up properly.** terrestris/geostyler-demo can only be set up properly, if we publish a release with the changes in this PR.

Closes #459 

This PR opens the gate for enabling a css preprocessor (#65)

This also solves the discussion in #526

### DIFF
--- a/.ci/update-gh-pages.sh
+++ b/.ci/update-gh-pages.sh
@@ -17,6 +17,11 @@ if [ $TRAVIS_BRANCH != "master" ]; then
   return 0;
 fi
 
+if ! [ -n "$TRAVIS_TAG" ]; then
+  # Only update when a tag was pushed / a new release was published
+  return 0;
+fi
+
 VERSION=$(node -pe "require('./package.json').version")
 
 ORIGINAL_AUTHOR_NAME=$(git show -s --format="%aN" $TRAVIS_COMMIT)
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -56,56 +52,68 @@
...
     "lodash": "4.17.11",
     "moment": "2.22.2",
-    "ol": "4.6.5",
-    "react": "16.6.0",
     "react-codemirror2": "5.1.0",
     "react-color": "2.14.1",
-    "react-dom": "16.6.0",
-    "react-rnd": "9.0.3",
-    "react-scripts-ts": "3.1.0",
+    "react-rnd": "9.0.2",
```
KaiVolland: Why did you downgrade `react-rnd`?

tslint.json
```diff
@@ -48,7 +45,6 @@
     "no-switch-case-fall-through": true,
     "no-trailing-whitespace": true,
     "no-unused-expression": true,
-    "no-unused-variable": true,
```
KaiVolland: Please re-add this rule

jansule: This rule is deprecated and is set in the compilerOptions of tsconfig.json.

package.json
jansule: Rebase error. Fixing with upcoming commit.

jansule:
@KaiVolland could you take a look at the last changes? I (re-)added the update-gh-pages.sh script and modified it a little. Now, it should create a directory for every upcoming version called `vVERSION` (e.g. `v0.18.1`) and copy the content of the latest version into the `latest` directory. I will manually add a new `index.html` to gh-pages that will redirect to `latest` directory. Thus, also older versions of styleguide will be available.

_Inspired by [@terrestris/base-util](https://github.com/terrestris/base-util/) and [@terrestris/base-util gh-pages](https://terrestris.github.io/base-util/latest/index.html)_

Issue #505, opened 2018-10-25
Performance


KaiVolland:
Due to the increasing size of the parsers and UIComponents we're loosing performance. We should investigate what causes the laggy behaviour and then fix it.

e.g. `shouldComponentUpdate` could help fixing problems with the UI performance.

KaiVolland:
Note: The `CodeEditor` component seems to slow down the application a lot.

jansule:
I made some simple performance checks. So far, we have two main problems here:
1. **every component will be re-rendered as soon as the style changes.** This also affects components that should actually not be updated, e.g. if there are changes in Rule1, Rule2 will also be updated, if there are changes in a symbolizer, filter will also be updated. React re-renders all subcomponents of a component that called setState, which in our case is the whole App.

2. `CodeEditor` **calls the active parser on each and every change.** E.g. if I replace a color value `#000000` to `#FFFFFF`, the style will be parsed 12 times! Once for every removal/addition of a char.

To counteract the first problem, we can use `shouldComponentUpdate()` at relevant parts to make sure that subcomponents only render if a relevant part of a style changed. Below images show the perfomance when removing and afterwards adding two characters in the `CodeEditor`, with and without `shouldComponentUpdate()`.

**Image1:** Changes in `CodeEditor` and collapsed `FilterTree`.  
_Upper Plot_: without `shouldComponentUpdate()`.  
_Lower Plot_: with  `shouldComponentUpdate()` on `FilterTree` and `ScaleDenominator`.
![image](https://user-images.githubusercontent.com/12186477/47788683-b889f980-dd12-11e8-8af5-d5d3223d6070.png)

**Image2:** Changes in `CodeEditor` and expanded `FilterTree`.  
_Upper Plot_: without `shouldComponentUpdate()`.  
_Lower Plot_: with  `shouldComponentUpdate()` on `FilterTree` and `ScaleDenominator`.
![image](https://user-images.githubusercontent.com/12186477/47789750-ba08f100-dd15-11e8-9f29-dea144aae9af.png)


It already shows clear performance improvements. The red ticks show, where the performance is very slow due to the bad performance of `CodeEditor`.

KaiVolland:
I thought about delaying the call of the change listeners for input fields as the `CodeEditor`.

This way we only call the `onStyleChange` function once the input is done.

e.g. for the `CodeEditor` which waits for 500ms before firing the change callback (untested code):

```diff
[...]

PR #589, opened 2018-11-15
Fix performance related bugs

### LINKED ISSUE #576: Some components do not re-render
Since we made the performance improvements, some components will not be re-rendered after changes were made (see example below).

**Before switching language**
image

**After switching language**
image

**After changing a property** (in this case, increasing radius)
image

jansule:
Some components did not update properly since our performance optimizations. Especially switching languages did not work properly, due to some parent components not having locales on their own, and thus, not updating child components. This PR fixes this.

Also added `build:dist` as default task to vscode to run it with `ctrl + shift + b` from within vscode.

Note: FilterTree is still quite slow in comparison to the other components. Relates to #505 

Closes #576

### DIFF
--- a/src/Component/Filter/FilterTree/FilterTree.tsx
+++ b/src/Component/Filter/FilterTree/FilterTree.tsx
@@ -12,6 +12,7 @@ import {
 const _get = require('lodash/get');
 const _set = require('lodash/set');
 const _isEqual = require('lodash/isEqual');
+const _cloneDeep = require('lodash/cloneDeep');
 
 const TreeNode = Tree.TreeNode;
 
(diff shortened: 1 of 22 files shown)


src/Component/Filter/FilterTree/FilterTree.tsx
```diff
@@ -323,7 +347,8 @@ export class FilterTree extends React.Component<FilterTreeProps, FilterTreeState
     } = this.props;
 
     let addedFilter: GsFilter ;
-    const newFilter: GsFilter = [...filter];
+    // const newFilter: GsFilter = [...filter];
```
hwbllmnn: Commented code

PR #607, opened 2018-11-20
Make rule renderer configurable

ahennr:
Make rule renderer configurable in `Rule.tsx`.
In addition, additional headers for POST request in `SLDRenderer` can be passed (e.g. CSRF token)

Depends on https://github.com/terrestris/geostyler/pull/605

Please review @terrestris/devs

### DIFF
--- a/src/Component/Rule/Rule.tsx
+++ b/src/Component/Rule/Rule.tsx
@@ -27,6 +27,7 @@ const _isEqual = require('lodash/isEqual');
 
 import './Rule.css';
 import en_US from '../../locale/en_US';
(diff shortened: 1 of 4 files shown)


src/Component/Rule/Rule.tsx
```diff
@@ -53,6 +54,7 @@ interface RuleDefaultProps {
   /** The data projection of example features */
   dataProjection: string;
   rendererType: 'SLD' | 'OpenLayers';
+  sldRendererProps: SLDRendererProps;
```
dnlkoch: Very minor (and probably not needed at the moment), but could you also provide an option to pass props for the `Renderer` (e.g. as `rendererProps`)? You decide.

PR #611, opened 2018-11-21
Pass futher props to FilterTree and FilterEditorWindow

ahennr:
Especially:

* FilterUiProps for attribute and operator name mappings
* The data definition and specification
* The renderer type

Depends on #610 

Plz review @terrestris/devs

### DIFF
--- a/src/Component/RuleTable/RuleTable.tsx
+++ b/src/Component/RuleTable/RuleTable.tsx
@@ -134,15 +134,18 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
     };
 
     return (
-      rendererType === 'SLD' ? <SLDRenderer
-      symbolizers={record.symbolizers}
-      onClick={onSymbolizerClick}
-      {...sldRendererProps}
-    />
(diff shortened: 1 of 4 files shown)


src/Component/RuleTable/RuleTable.tsx
```diff
@@ -114,13 +123,25 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
...
+    };
+
     return (
-      <Renderer
+      rendererType === 'SLD' ? <SLDRenderer
+      symbolizers={record.symbolizers}
+      onClick={onSymbolizerClick}
+      {...sldRendererProps}
+    />
+      : <Renderer
```
jansule: Minor: Could you add brackets arount the Element, to make sure the whole element will be read?
Example: 
```
rendererType === 'SLD' ? (<SldRenderer
                             ...props
                           />) : ....
```

ahennr: :white_check_mark: , see fbe63e9

PR #613, opened 2018-11-21
Removes duplicate entry in locale de

ahennr:
Title says it all.

Fixes bug in IE11 that doesn't allow duplicate entries in strict mode.


marcjansen:
Thanks for the fix. Can we enhance our eslint rules to catch these in the future earlier?

KaiVolland:
@marcjansen Should be fixed by transforming the locale files from 'de_de.js' to 'de_de.ts'

PR #619, opened 2018-11-22
Replace ButtonGroup with FooterMenu in RuleTable

jansule:
Added property `footer` to RuleTable. This passes through a custom footer to antd Table. Removed ButtonGroup below RuleTable and added a Menu into the footer, instead.

image

### DIFF
--- a/src/Component/Style/Style.tsx
+++ b/src/Component/Style/Style.tsx
@@ -204,7 +204,7 @@ export class Style extends React.Component<StyleProps, StyleState> {
     });
   }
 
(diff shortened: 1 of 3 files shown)


src/Component/Style/Style.tsx
```diff
@@ -206,6 +206,12 @@ export class Style extends React.Component<StyleProps, StyleState> {
 
   onMultiEdit = (param: any) => {
```
KaiVolland: Rename please.

PR #622, opened 2018-11-23
Set default well known name

hwbllmnn:
The styler would crash if the user would classify and not choose something in the wellKnownName dropdown.


KaiVolland:
Good for know. We should make all componets more failsafe. I created an issue therefore: #623

PR #632, opened 2018-11-26
Restrict multi-edits in RuleTable

jansule:
The multi-edit submenu in the footer of RuleTable will now be disabled if the number of selected rows is `<=1`.

Also, only those menu items of multi-edit are enabled that kann actually be changed for all symbolizers in the current selection. E.g. if a symbolizer of kind 'Line' and a symbolizer of kind 'Mark' are selected, the `size` and `symbol` items are disabled.

image

image

### DIFF
--- a/src/Component/Style/Style.tsx
+++ b/src/Component/Style/Style.tsx
@@ -276,6 +276,13 @@ export class Style extends React.Component<StyleProps, StyleState> {
     this.setState({ruleGeneratorWindowVisible: false});
   }
 
+  /**
+   * Checks if a specific menu item of multi-edit menu should be disabled.
+   *
+   * @param name Name of menu item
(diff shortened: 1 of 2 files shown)


src/Component/Style/Style.tsx
```diff
@@ -276,6 +276,49 @@ export class Style extends React.Component<StyleProps, StyleState> {
     this.setState({ruleGeneratorWindowVisible: false});
   }
 
+  disableMenu = (name: string, rowKeys: number[]): boolean => {
```
KaiVolland: A small documentation would be nice.

jansule: Done in https://github.com/terrestris/geostyler/pull/632/commits/efabb8fe2e75965d190e9cf62a4e9f1016d1cd31

PR #639, opened 2018-11-27
Introduce Columns Amount and Duplicates to RuleTable

jansule:
* Columns `Amount` and `Duplicates` were added to RuleTable
* Added functions to calculate values for those columns
* Set cursor style to default in `Filter` column to emphasize that this input  is not editable

_shows `-` if no data is given_
image

image

image

### DIFF
--- a/src/Util/FilterUtil.ts
+++ b/src/Util/FilterUtil.ts
@@ -2,8 +2,17 @@ import {
   Filter,
   Operator,
   CombinationOperator,
-  NegationOpertaor
+  NegationOpertaor,
+  Rule
 } from 'geostyler-style';
+
+import {
+  Data
+} from 'geostyler-data';
+
+const _get = require('lodash/get');
+const _isEqual = require('lodash/isEqual');
+const _cloneDeep = require('lodash/cloneDeep');
 
 /**
  * @class SymbolizerUtil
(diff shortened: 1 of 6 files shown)


src/Util/FilterUtil.ts
```diff
@@ -2,9 +2,18 @@ import {
   Filter,
   Operator,
   CombinationOperator,
-  NegationOpertaor
+  NegationOpertaor,
```
hwbllmnn: Not really your change, but this is definitely a typo

src/Util/FilterUtil.ts
```diff
@@ -71,6 +80,182 @@ class FilterUtil {
...
+          if (typeof prop === typeof filter[2]) {
+            matchesFilter = prop >= filter[2];
+          }
+          break;
+        default:
+          throw new Error(`Cannot parse Filter. Unknown comparison operator.`);
+      }
+    }
+    return matchesFilter;
+  }
```
hwbllmnn: It would be nice if you could split this up, this function is nearly 100 lines

marcjansen:
Are duplicates correctly calculated? Looks a bit odd... but I might be wrong. The column titles could be enhanced e.g. Sigma for sum and some other sign for dupes? with tooltips giving the full explanation?

src/Util/FilterUtil.ts
jansule: True, this needs to be fixed at first in geostyler-style and then on all other packages using this type

src/Util/FilterUtil.ts
jansule: Done in https://github.com/terrestris/geostyler/pull/639/commits/788f981208c617a5791e71586aeb4629198cdbf5

src/Util/FilterUtil.ts
marcjansen: AFAICT I fixed this in a PR...

marcjansen: https://github.com/terrestris/geostyler-style/pull/80

jansule:
@marcjansen 
> Are duplicates correctly calculated? Looks a bit odd... but I might be wrong.

Yes, they are. The example in the screenshot might confuse a little.

> The column titles could be enhanced e.g. Sigma for sum and some other sign for dupes? with tooltips giving the full explanation?

Done in https://github.com/terrestris/geostyler/pull/639/commits/112934b78e89d7a11b2dbbab61bd40799feb8bbe

image

src/Util/FilterUtil.ts
```diff
@@ -71,6 +80,182 @@ class FilterUtil {
...
+          matchesFilter = !FilterUtil.featureMatchesFilter(filter[1], feature);
+          break;
+        default:
+          throw new Error(`Cannot parse Filter. Unknown combination or negation operator.`);
+      }
+    } else {
+      const prop: any = _get(feature, 'properties[' + filter[1] + ']');
+      switch (filter[0]) {
+        case '==':
+          // tslint:disable-next-line
```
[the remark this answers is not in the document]
jansule: I used the equality operator (`==`) instead of the identity operator (`===`) there. Got a little confused after yesterday's discussion about parsing properties in sld.
Solved in https://github.com/terrestris/geostyler/pull/639/commits/1300ace7450b8804767bd6964d5ad38a18f83e9f

jansule:
> Please don't use `"copy"` as it suggests that you can copy something here.

Done in https://github.com/terrestris/geostyler/pull/639/commits/e467152f125709cfb52e7e9688bd9b8941f6af50

image

PR #641, opened 2018-11-27
Use whatwg-fetch for browser builds

ahennr:
Since older browsers don't support the `fetch` API, this PR introduces `whatwg-fetch` to browser-build config.

Please review

### DIFF
--- a/browser-build.config.js
+++ b/browser-build.config.js
@@ -5,7 +5,11 @@ const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 require("@babel/polyfill");
 
 module.exports = {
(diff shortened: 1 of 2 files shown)


KaiVolland:
Shouldn't we add this to the `webpack.common.config.js` to?

ahennr:
Currently, `webpack.common.config.js` is used for styleguide, which doesn't work in IE11 currently. Beside, the geostyler demo application doesn't work in IE11 as well. IMHO, these two issues should be fixed in an upcoming PR, shouldn't it @KaiVolland @jansule ?

ahennr:
See also: https://github.com/terrestris/geostyler/issues/642

PR #645, opened 2018-11-28
[WIP] Fix gh-pages (travis-ci.com)

jansule:
Using new travis configs instead of our own script. gh-pages will only be updated if a tag was pushed. Creates a folder for the current version and copies it's content into `latest` folder.

gh-pages will have following structure
```
root/
----v0.22.0/
----v0.22.1/
----.../
----latest/
----favicon.ico
----index.html
```

index.html always redirects to `latest/`. Older versions will be accessible via `terrestris.github.io/geostyler/vVERSION/`


KaiVolland:
We have to wait until the project is migrated to travis.com

KaiVolland:
>We have to wait until the project is migrated to travis-ci.com

Seems like nothing happens here. We should try to fix gh-pages for the current travis-ci.org setup ...

KaiVolland:
Should be fixed by #945

PR #646, opened 2018-11-28
Always compare as float for > and < filters

hwbllmnn:
This fixes the comparison when the filter was parsed (at least) from SLD. In that case all literals will be stored internally as strings, thus previously all comparisons would fail.

This also forces (in-)equality comparisons as strings.

@terrestris/devs @chrismayer Do you have any thoughts on this? IMHO this approach makes the most sense:

* comparing (in-)equality will work perfectly for numbers and strings alike when done as strings
* comparing with `PropertyIsLike` filters makes only really sense in case all operands are strings anyway
* all other comparisons make only sense when using numbers (although the edge case of classifying with the beginning letter might come around some day)

### DIFF
--- a/src/Util/FilterUtil.ts
+++ b/src/Util/FilterUtil.ts
@@ -114,28 +114,29 @@ class FilterUtil {
    * Handle simple filters, i.e. non-nested filters.
    */
   static handleSimpleFilter = (filter: Filter, feature: any): boolean => {
-    const prop: any = _get(feature, 'properties[' + filter[1] + ']');
+    const featureValue: any = _get(feature, 'properties[' + filter[1] + ']');
+    const filterValue = filter[2];
     switch (filter[0]) {
       case '==':
-        return (('' + prop) === ('' + filter[2]));
+        return (('' + featureValue) === ('' + filterValue));
       case '*=':
-        if (prop && filter[2].length > prop.length) {
+        if (featureValue && filterValue.length > featureValue.length) {
           return false;
-        } else if (prop) {
-          return (prop.indexOf(filter[2]) !== -1);
+        } else if (featureValue) {
+          return (featureValue.indexOf(filterValue) !== -1);
         } else {
           return false;
         }
       case '!=':
-        return (('' + prop) !== ('' + filter[2]));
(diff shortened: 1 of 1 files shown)


src/Util/FilterUtil.ts
```diff
@@ -117,7 +117,7 @@ class FilterUtil {
     const prop: any = _get(feature, 'properties[' + filter[1] + ']');
```
KaiVolland: Even if not related to this PR. This should be renamed to `featureValue`. Additonaly
```js
const filterValue = filter[2];
```
would add some clarification...

src/Util/FilterUtil.ts
```diff
@@ -117,7 +117,7 @@ class FilterUtil {
...
     switch (filter[0]) {
       case '==':
-        return (prop === filter[2]);
+        return (('' + prop) === ('' + filter[2]));
```
KaiVolland: Nice approach but i'm afraid we won't fit any cases of string-number comparsion... (don't know if that is what we want)

e.g When `featureValue` is `'2.00'` and `filterValue` is `2`:

```javascript
'' + '2.00' === '' + 2 // false
```

hwbllmnn: True, but then again the previous approach would also lead to `false` here.

src/Util/FilterUtil.ts
hwbllmnn: Done in f9b4acb

chrismayer:
Even this is merged already a little thought in this:

Maybe a comparison with `== ` would be an option. At least numbers and their string representation would match. Or we introduce a 'comparisonMode' property whose value (e.g. 'STRICT' and 'SMART') influences the comparison mechanism. As said just some thoughts and still workarounds IMHO. The root of all evil ist that SLD is not aware of data types :-(

hwbllmnn:
That's a neat idea, I've opened [an issue](#655) for this, thanks!

chrismayer:
Thanks @hwbllmnn for opening the issue. The link above seems a bit mixed up, so -> #655

PR #657, opened 2018-11-30
Adds "Copy to Clipboard" to CodeEditor

KaiVolland:
This adds a "Copy to Clipboard" Button to the `CodeEditor` component coming with a `showCopyButton` property.

`showSaveButton` and `showCopyButton` props are now set to `false` per default.

localhost_3000_ 24

This also changes the fileending for saved `SLD` files from `[Style.name].xml` to `[Style.name].sld`

### DIFF
--- a/src/Component/CodeEditor/CodeEditor.tsx
+++ b/src/Component/CodeEditor/CodeEditor.tsx
@@ -15,7 +15,11 @@ import {
 
 import './CodeEditor.css';
 
-import { Select, Button } from 'antd';
+import {
+  Button,
+  message,
+  Select
+} from 'antd';
 const Option = Select.Option;
 
 import {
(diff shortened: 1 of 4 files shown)


src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -237,20 +244,56 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
       style
     } = this.props;
     if (style) {
+      let fileName = style.name;
       let type = 'application/json;charset=utf-8';
       if (activeParser && activeParser.title === 'SLD Style Parser') {
         type = 'text/xml;charset=utf-8';
+        fileName += '.sld';
```
hwbllmnn: Suggestion:

```suggestion
        if (!fileName.endsWith('.sld')) {
            fileName += '.sld';
        }
```

KaiVolland: As the filename is just `style.name` it is very unlikely that it ends withs '.sld'...

PR #683, opened 2018-12-10
Improves performance of duplicates detection

hwbllmnn:
This now only calculates the amount and duplicates if the data or a filter change. Also, the duplicates detection now uses the results of calculating the matches while determining the count, so the features are only matched once against the filters.

### DIFF
--- a/src/Component/RuleTable/RuleTable.tsx
+++ b/src/Component/RuleTable/RuleTable.tsx
@@ -79,10 +79,8 @@ interface RuleTableState {
   hasError: boolean;
   data: GsData;
   rules: GsRule[];
-  countsAndDuplicates: {
-    counts: number[],
-    duplicates: number[]
-  };
+  counts: number[];
+  duplicates: number[];
 }
 
 export interface RuleRecord extends GsRule {
@@ -104,7 +102,8 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
       hasError: false,
       data: undefined,
       rules: undefined,
-      countsAndDuplicates: undefined
+      counts: undefined,
+      duplicates: undefined
     };
   }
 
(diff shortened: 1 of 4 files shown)


src/Component/RuleTable/RuleTable.tsx
```diff
@@ -77,6 +77,12 @@ interface RuleTableState {
   filterEditorVisible: boolean;
   filterEditorPosition: DOMRect;
   hasError: boolean;
+  data: GsData;
+  rules: GsRule[];
+  countsAndDuplicates: {
```
KaiVolland: I'd prefer to have this split in just `counts` and `dupliactes`.

src/Component/RuleTable/RuleTable.tsx
```diff
@@ -104,6 +113,85 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
...
+      // make sure to update state when checks/calculation fails
+    }
+    return {
+      data: nextProps.data,
+      rules: nextProps.rules,
+      countsAndDuplicates
+    };
+  }
+
+  static calculateDuplicates(matches: any[][]): number[] {
```
KaiVolland: I think this could be moved to the Util.

src/Component/RuleTable/RuleTable.tsx
```diff
@@ -104,6 +113,85 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
...
+          });
+        }
+      });
+      duplicates.push(counter);
+    });
+
+    return duplicates;
+  }
+
+  static calculateCountAndDuplicates(rules: GsRule[], data: GsData): {
```
KaiVolland: I think this could be moved to the Util.

src/Component/RuleTable/RuleTable.tsx
hwbllmnn: Done in a49449f

src/Component/RuleTable/RuleTable.tsx
hwbllmnn: Done in a49449f

src/Component/RuleTable/RuleTable.tsx
hwbllmnn: Done in a49449f

PR #692, opened 2018-12-13
Introduces the export of typescript declarations

### LINKED ISSUE #684: Add Declartion file
### Description:

We should introduce a declaration file (`index.d.ts`) so users of the GeoStyler can benefit from its typing.

Usually you just need to add `"declaration": true` to your tsconfig.json. And typescript will create it during build.

Currently there is a conflict with the hidden `Wrap` class in the `LocaleWrapper` which doesn't allow us to do so. We should have a look at this.

KaiVolland:
This introduces the export of TypeScript declarations for the GeoStyler.
In other words the `geostyler` package will include a `index.d.ts` to benefit from TypeScript when using it.

- All *.js files are replaced with *.ts equivalents (locales and index.js)
- Includes some minor changes to the `SLDRendererProps` and `LocaleWrapper`

Closes #684


package.json
```diff
@@ -35,7 +36,6 @@
     "release": "np --no-yarn && git push https://github.com/terrestris/geostyler.git master --tags",
     "styleguide": "styleguidist server",
     "start:styleguide": "styleguidist server",
-    "start:dev": "webpack --mode=development --watch --config dev-build.config.js",
```
hwbllmnn: Is there another way to have a dev build continually built? This is really handy when developing with non es6 projects.

KaiVolland: I was quite sure that this was an old script so i removed it... I'll have a look at it, and bring it back maybe.

PR #700, opened 2018-12-17
Ol5 upgrade

annarieger:
### BREAKING CHANGE 

This PR can be considered as basic preparation for the future transition to the `ol5`. Changes are expected after TypeScript definitions for the `ol5` release are published (s. also https://github.com/openlayers/openlayers/issues/8120)

**Notes**: 

1. Due to this upgrade your parent package/application won't work with `ol` version older than 5.x To restore the functionality please refer to the official [upgrade guide](https://github.com/openlayers/openlayers/releases/tag/v5.0.0).

2. A `custom.d.ts` file containing currently missing ol5 type declarations is added. After `@types/ol5` is done this file is very probably to be removed again as obsolete.

3. Currently all `ol` related types were replaced by very common `any`  type (s. 
32684a3) - this commit is very likely to be reverted after `types/ol5` is avaliable.

Depends on https://github.com/terrestris/geostyler-openlayers-parser/pull/77 (must be merged before, otherwise checks won't never get green)

Please review @dnlkoch @KaiVolland


custom.d.ts
```diff
@@ -0,0 +1,13 @@
...
+declare module 'ol/layer/Vector';
+declare module 'ol/source/OSM';
+declare module 'ol/source/Vector';
+// will be declared in geostyler-openlayers-parser
+// declare module 'ol/Feature';
+declare module 'ol/Map';
+declare module 'ol/View';
+declare module 'ol/geom/LineString';
+// will be declared in geostyler-openlayers-parser
+// declare module 'ol/geom/Point';
```
dnlkoch: I had to include both declarations to get things running. Maybe this was related to a `npm link` to the `geostyler-openlayers-parser` in your local checkout?

annarieger: Fixed in bf54595

annarieger:
> Why is the coverage drop so big?
@marcjansen, coverage is not changed. Possibly the previously shown gap of almost 10% can be explained with some failed tests..

annarieger:
So, thanks for the reviews, I'll give ol5 a try now..

PR #715, opened 2018-12-21
RuleTable: Make visibility of columns for duplicates and amount configurable

ahennr:
Title says it all.


ahennr:
Thx for review.

PR #728, opened 2019-01-02
Adds ids to features if not present

KaiVolland:
This fixes a bug in duplicates calculation.

As the calculation of duplicates relies on `feature.id` we use the index inside the `FeatureCollection` if the feature has no id.

### DIFF
--- a/src/Util/FilterUtil.ts
+++ b/src/Util/FilterUtil.ts
@@ -218,6 +218,15 @@ class FilterUtil {
       counts: [],
       duplicates: []
     };
+
+    // Add id to feature if missing
+    data.exampleFeatures.features = data.exampleFeatures.features.map((feature, idx) => {
+      if (!feature.id) {
+        feature.id = idx;
+      }
+      return feature;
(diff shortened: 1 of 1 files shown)


src/Util/FilterUtil.ts
```diff
@@ -218,6 +218,15 @@ class FilterUtil {
       counts: [],
       duplicates: []
     };
+
+    // Add id to feature if missing
+    data.exampleFeatures.features = data.exampleFeatures.features.map((feature, idx) => {
+      if (!feature.id) {
+        feature.id = idx;
```
ahennr: [Lodash's](https://lodash.com/docs/4.17.11#uniqueId) `uniqueId` function could help to generate unique ids (e.g. with feature type as prefix), what do you think?

KaiVolland: That was my first thought, but i think this is should be faster and fit our needs. I think there will be no FeatureCollection with "mixed content" where some features have ids and some not.

PR #737, opened 2019-01-10
Allows nominal classification for all field types

hwbllmnn:
This removes the restriction of allowing nominal classification only for string fields. Perhaps this does not always make sense (e.g. for measure values), but in case of e.g. numeric codes such as zip codes it might be useful.

@KaiVolland Please review.


KaiVolland:
You're right. Did you test it with larger datasets?

hwbllmnn:
I tested it with around 1000 features, the classification is almost instant.

PR #799, opened 2019-02-06
Adds flag to have custom color ramps plus the brewer ones

hwbllmnn:
Adds a flag to the styler component to merge the custom color ramps with the chroma-js brewer ramps.

### DIFF
--- a/src/Component/Style/Style.tsx
+++ b/src/Component/Style/Style.tsx
@@ -79,6 +79,7 @@ export interface StyleProps extends Partial<StyleDefaultProps> {
   colorRamps?: {
     [name: string]: string[]
   };
+  useBrewerColorRamps?: boolean;
   colorSpaces?: (InterpolationMode)[];
 }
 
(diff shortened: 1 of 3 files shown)


src/Component/Style/Style.tsx
```diff
@@ -79,6 +79,7 @@ export interface StyleProps extends Partial<StyleDefaultProps> {
   colorRamps?: {
     [name: string]: string[]
   };
+  useBrewerColorRamps?: boolean;
```
KaiVolland: Should be moved to `StyleDefaultProps` and added with true to `defaultProps`.

hwbllmnn: IMHO a default value of `true` does not really make sense, since setting it to `true` means that you also need to configure custom color ramps (which by default are not set).

KaiVolland: Ok so i didn't get it. I'll read it again.

KaiVolland: Thought `Style` was the final component... Just forget it and merge. :)

PR #815, opened 2019-02-13
Introduces jest v24

KaiVolland:
This introduces jest in v24.

Jest 24 uses babel 7 which supports TypeScript. Therefore we can remove ts-jest and use some babel presets and plugins instead.

@terrestris/devs please review

closes #810 
closes #814


jest.config.js
```diff
@@ -5,32 +5,29 @@ module.exports = {
...
+  ],
   "testMatch": [
     "<rootDir>/src/**/__tests__/**/*.(j|t)s?(x)",
     "<rootDir>/src/**/?(*.)(spec|test).(j|t)s?(x)"
   ],
   "testURL": "http://localhost",
   "transform": {
     "^.+\\.jsx?$": "<rootDir>/node_modules/babel-jest",
-    "^.+\\.tsx?$": "ts-jest",
+    "^.+\\.tsx?$": "<rootDir>/node_modules/babel-jest"
```
dnlkoch: Or something like:
```
"^.+\\.(tsx|jsx)?$": "<rootDir>/node_modules/babel-jest"
```

jest/setup.js
```diff
@@ -1,4 +1,5 @@
+require('@babel/polyfill');
```
dnlkoch: Or `import '@babel/polyfill';`

KaiVolland:
Coverage calculation seems to be different. :disappointed: 
Seems like we need to put some effort into tests again.

PR #816, opened 2019-02-13
More tests

KaiVolland:
This adds a some tests to reincrease the coverage a little. Still a lot of untested lines…

### DIFF
--- a/src/Util/HTTPUtil.spec.ts
+++ b/src/Util/HTTPUtil.spec.ts
@@ -1,10 +1,37 @@
-import HTTPUtil from './HTTPUtil';
+import HTTPUtil, { PostOptions } from './HTTPUtil';
 
 describe('HTTPUtil', () => {
 
   describe('post', () => {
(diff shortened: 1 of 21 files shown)


src/Util/HTTPUtil.spec.ts
```diff
@@ -1,11 +1,38 @@
...
+      fetch = jest.fn();
+      const headers = new Headers();
+      headers.set('Content-Type', 'application/x-www-form-urlencoded');
+      headers.set('superheader', 'header-abc');
+      // const expectedParams = {
+      //   headers,
+      //   method: 'POST',
+      //   credentials: 'same-origin',
+      //   body: new URLSearchParams({})
+      // };
```
jansule: Minor: Please remove outcommented code

KaiVolland: This code belongs to the comment below. So i'd like to keep it.

PR #862, opened 2019-02-26
Ignore scale denominators when calculating duplicates

hwbllmnn:
@KaiVolland Please review.


KaiVolland:
I don't think we should ignore the scale. It was introduced in #732 and makes sense to me.
If you got the same rule for two different scales the matched feature is not covered twice.
So in my opinion its not a duplicate.

KaiVolland:
... seems like QGIS also ignores the Scale. I still don't like it but this way but it should be close to QGIS so merge it please.

PR #868, opened 2019-02-26
Updates styleguidist to v9

KaiVolland:
This updates stylegudist to version 9.
It also modifies the examples as needed and adds missing packages.

@terrestris/devs please review

closes #866
closes #847


styleguide.config.js
```diff
@@ -47,5 +48,11 @@ module.exports = {
       name: 'UploadButton',
       components: 'src/Component/UploadButton/**/*.tsx'
     }]
-  }]
+  }],
+  // moduleAliases: {
+  //   'antd': path.resolve(__dirname, 'node_modules/antd')
+  // },
```
dnlkoch: Can be removed.

package.json
```diff
@@ -92,6 +92,7 @@
     "case-sensitive-paths-webpack-plugin": "2.2.0",
     "copyfiles": "2.1.0",
     "coveralls": "3.0.3",
+    "css-loader": "2.1.0",
```
dnlkoch: Is this really needed for the antd css require in `webpack.config.json`?

package.json
```diff
@@ -108,9 +109,10 @@
     "react": "16.8.3",
     "react-docgen-typescript": "1.12.3",
     "react-dom": "16.8.3",
-    "react-styleguidist": "8.0.6",
+    "react-styleguidist": "9.0.1",
     "rimraf": "2.6.3",
     "source-map-loader": "0.2.4",
+    "style-loader": "0.23.1",
```
dnlkoch: See comment above.

src/Component/RuleGenerator/RuleGenerator.example.md
```diff
@@ -1,8 +1,7 @@
 This demonstrates the use of `RuleGenerator`.
 
 ```jsx
-const React = require('react');
-const { RuleGenerator } = require('../../index');
```
dnlkoch: Even if not needed for styleguide itself, I would suggest to import all related classes for better real world appliance of the examples.

KaiVolland: I had to remove it, as i got 'already defined' errors when leaving the import in the example.

package.json
KaiVolland: Yes. 'module not found' errors where thrown before.

package.json
KaiVolland: Yes. 'module not found' errors where thrown before.

PR #881, opened 2019-03-05
Constrain the symbolizer window to the body

hwbllmnn:
Without this, it's possible to move the window partly out of the browser, possibly being stuck without a moving handle.


KaiVolland:
Please make this configurable.

hwbllmnn:
Ok to do it like this? Or do we really need this to be configurable the `Style.tsx`?

KaiVolland:
> Ok to do it like this? Or do we really need this to be configurable in the `Style.tsx`?
I'm fine with the PR. We might consider to introduce a prop to the `Style` later.

PR #893, opened 2019-03-11
Allow parser instances instead of constructor functions

hwbllmnn:
Allows parser instances to be specified in the parser list in order to be able to construct parsers with options.

### DIFF
--- a/src/Component/CodeEditor/CodeEditor.tsx
+++ b/src/Component/CodeEditor/CodeEditor.tsx
@@ -24,7 +24,8 @@ const Option = Select.Option;
 
 import {
   Style as GsStyle,
-  StyleParserConstructable as GsStyleParserConstructable
+  StyleParserConstructable as GsStyleParserConstructable,
+  StyleParser
 } from 'geostyler-style';
 
 const _isEqual = require('lodash/isEqual');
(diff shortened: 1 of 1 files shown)


src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -126,23 +127,19 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
...
-    const {
-      activeParser
-    } = this.state;
+    const activeParser: any = this.state.activeParser;
     return new Promise((resolve, reject) => {
       if (activeParser) {
-        const StyleParser = activeParser;
-        const parserInstance = new StyleParser();
+        const StyleParserClass = activeParser;
+        const parserInstance = activeParser instanceof Function ? new StyleParserClass() : activeParser;
```
KaiVolland: This seems to be kind of a loose check. But not blocking actually.

src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -151,13 +148,11 @@ export class CodeEditor extends React.Component<CodeEditorProps, CodeEditorState
...
```
KaiVolland: This seems to be kind of a loose check. But not blocking actually.

src/Component/CodeEditor/CodeEditor.tsx
hwbllmnn: It is, but unfortunately the `instanceof StyleParser` cannot be used since it is an interface type.

src/Component/CodeEditor/CodeEditor.tsx
hwbllmnn: See above.

PR #926, opened 2019-04-01
Fix crash if text symbolizer has no label set

hwbllmnn:


hwbllmnn:
Using `value` instead of `defaultValue` causes the label field not being editable any more. BTW, do you have any idea why CI is failing? The build works just fine for me locally.

KaiVolland:
> Using `value` instead of `defaultValue` causes the label field not being editable any more. BTW, do you have any idea why CI is failing? The build works just fine for me locally.

Hmm that's strange. Works like this for every other antd component.
> BTW, do you have any idea why CI is failing? The build works just fine for me locally.
I'll have a look.

KaiVolland:
Maybe the recent package updates have caused another occasion of mixed `@types` packages...

KaiVolland:
The types need to stay in the `dependencies` so `"types": "dist/index.d.ts"` brings all the types to the user.

hwbllmnn:
It seems the failures were caused by the `@types` being split between `devDependencies` and `dependencies`. Moving them all to `dependencies` fixes the problem.

KaiVolland:
Still strange. We should keep an eye on this.

Could you try to keep the definietly testing realted packages in the `devDependencies`?
```js
"@types/enzyme": "3.9.1",
"@types/jest": "24.0.11",
"@types/jest-diff": "20.0.1",
"@types/node": "11.11.6"
```
If you still got issues, feel free to merge anyway. We should have an eye on this, especially regarding to new antd releases.

hwbllmnn:
Squash-merged.

PR #934, opened 2019-04-05
Updates packages and adapt code

KaiVolland:
Updates several dependencies and removes `StyleParserConstrutable` from `SLDRenderer`.


KaiVolland:
The `DataLoader` currently still relies on the Constructables so this will follow in a later PR.

PR #1000, opened 2019-04-26
Add data to Renderer

KaiVolland:
With this change the internal data gets passed to the `Renderer` so we can preview template strings with curly brackets: e.g:

image

This also makes the `rasterLayer` prop of the `SLDRenderer` optional.

### DIFF
--- a/src/Component/Symbolizer/Renderer/Renderer.tsx
+++ b/src/Component/Symbolizer/Renderer/Renderer.tsx
@@ -17,7 +17,7 @@ import './Renderer.css';
 
 import 'ol/ol.css';
 import './Renderer.css';
-import { Data, VectorData } from 'geostyler-data';
+import { Data } from 'geostyler-data';
 
 const _isEqual = require('lodash/isEqual');
 const _get = require('lodash/get');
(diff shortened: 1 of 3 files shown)


src/Component/Symbolizer/Renderer/Renderer.tsx
```diff
@@ -80,10 +82,14 @@ export class Renderer extends React.Component<RendererProps> {
   }
 
   updateFeature() {
+    const vectorData = this.props.data as VectorData;
```
jansule: [DataUtil](https://github.com/terrestris/geostyler/blob/master/src/Util/DataUtil.ts) provides functions for checking the type of data (vector/raster). You might want to use it here.

src/Component/Symbolizer/SLDRenderer/SLDRenderer.tsx
```diff
@@ -16,7 +16,7 @@ interface SLDRendererDefaultProps {
 export interface SLDRendererAdditonalProps extends Partial<SLDRendererDefaultProps> {
   wmsBaseUrl: string;
   layer: string;
-  rasterLayer: string;
+  rasterLayer?: string;
```
jansule: Is this intended or just an artefact? If intended, it should be checked [here](https://github.com/terrestris/geostyler/pull/1000/files#diff-8a2ec6e31ba513cd599fb0bb89da0ebbR118) if `rasterLayer` is defined.

Issue #831, opened 2019-02-15
Highligh Rule SymbolizerEditorWindow


jansule:
### Description:

When working with many rules, one might get confused to which rule the currently opened SymbolizerEditorWindow belongs. This problem could be counteracted via highlighting the rule that is currently being edited.

mholthausen:
See PR #1073

PR #1073, opened 2019-05-29
Highlighting the rule that is currently being edited via the checkbox

mholthausen:
Handles the request from Issue #831 

When opening the SymbolyzerEditorWindow the current rule is marked via the checkbox and is removed again when closing it.

### DIFF
--- a/src/Component/RuleTable/RuleTable.tsx
+++ b/src/Component/RuleTable/RuleTable.tsx
@@ -197,8 +197,10 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
     const {
       rowSelection
     } = this.props;
-    const onChange = rowSelection.onChange;
-    onChange([record.key], null);
+    if(typeof(rowSelection) != 'undefined' && rowSelection != null) {
+      const onChange = rowSelection.onChange;
+      onChange([record.key], null);
+    }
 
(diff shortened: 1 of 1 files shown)


src/Component/RuleTable/RuleTable.tsx
```diff
@@ -193,6 +193,12 @@ export class RuleTable extends React.Component<RuleTableProps, RuleTableState> {
   }
 
   onSymbolizerClick = (record: RuleRecord, symbolizerEditorPosition: DOMRect) => {
+    const {
+      rowSelection
+    } = this.props;
+    const onChange = rowSelection.onChange;
```
dnlkoch: Tests are failing, because `rowSelection` can be `null` right here. Please add a check.

KaiVolland:
Hey @mholthausen do you think you can finish this in the next couple of days? Otherwise i may try to continue your work from here on …

mholthausen:
@KaiVolland I have included the request mentioned above now, all tests are green.

KaiVolland:
Wouldn't it be better to just change the background color of the rule which is currently edited. Doing a real select via checkbox might have unwanted implications.

KaiVolland:
I'll close this for cleanup reasons.

PR #1112, opened 2019-06-18
Reintroduces package-lock and ^ versions

KaiVolland:
This reintroduces the `package-lock.json` and `^`-versions to reduce the amount of Greenkeeper PullRequests and the therefore upcoming effort.

Greenkeeper will not open PullRequests for BugFix releases anymore.

~~As we experienced compatiblity difficulties with feature releases in the past i'd prefer `~` over `^` even if it should work, too.~~ Seems as Greenkeeper is smart enough to handle this.

closes #1111


marcjansen:
> This reintroduces the package-lock.json and ~-versions …

Yes, `+1`

> … to reduce the amount of Greenkeeper PullRequests and the therefore upcoming effort.

Why would this reduce this?

> Greenkeeper will not open PullRequests for BugFix releases anymore.

Are you sure???. Than this is a `-1`. This is what we want GK for.

> As we experienced compatiblity difficulties with feature releases in the past i'd prefer ~ over ^ even if it should work, too.

Yes, `+1, even though this shouldn't happen and I think if it does this says something about the quality of the dependency itself. I want all the PRs and be sure what we do/what'll happen if we merge.

KaiVolland:
> Are you sure???. Than this is a -1. This is what we want GK for.

GK wont open PRs for bug-fix releases as every `npm i` will install the latest bug-fix release. At least this is what i expect to happen. :)

You can configure automated PRs for security relevant package updates [here](https://github.com/terrestris/geostyler/network/alerts). I think this could be usefull if we merge this PR.

KaiVolland:
Actually GK opens the PR anyways but deletes it if everything still runs as aspected. See https://greenkeeper.io/docs.html#greenkeeper-step-by-step

So best of both. CI is still tested but no PR will be opened! Actually GK seems to be so smart, that we could go for the `^` releases too. @terrestris/devs opinions?

weskamm:
:+1: from me too as we will also get that useful "automatic issue" by GK if an update fails in the tests and it GK retries when newer versions come out.

Issue #1134, opened 2019-07-11
broken link on documentation


pvgenuchten:
### Description:
On the documentation [index page](https://terrestris.github.io/geostyler/latest/index.html#introduction) the link to [geostyler docs](https://terrestris.github.io/geostyler/styleguide/) is broken

![image](https://user-images.githubusercontent.com/299829/61030467-b6a62900-a3bd-11e9-8c83-98249aa72b56.png)

KaiVolland:
@pvgenuchten Thx for contributing.

Seems like the real problem is, that the documentation didn't update since: https://github.com/terrestris/geostyler/commit/4c7d758064b5e428a45208785d1d87e2269301bb

I'll have a look at it.

chrismayer:
Hi @KaiVolland, are there any news on this?

KaiVolland:
I tried to fix this with #1138. But it didn't help. Seems like there is still something wrong with the CI.

chrismayer:
Thanks for the feedback @KaiVolland. Please ping me if I can support you somehow.

dnlkoch:
Automatic update of the `gh-pages` has been fixes by #1155. I'll close this issue as the described bug should be resolved by now. 

Please note #1158.

PR #1143, opened 2019-07-22
1134 gh pages script

chrismayer:
This fixes the detection of Pull Requests in the `update-gh-pages.sh` script and adds some more logs to get a clearer picture why the script stops.

Related to #1134.

### DIFF
--- a/.ci/update-gh-pages.sh
+++ b/.ci/update-gh-pages.sh
@@ -9,16 +9,19 @@ fi
 
 if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
   # Dont build anything for PR requests, only for merges.
+  echo "Skip script since it processes a Pull Request, not a merge."
   return 0;
 fi
 
 if [ "$TRAVIS_BRANCH" != "master" ] && [ -z "$TRAVIS_TAG" ]; then
   # Only update when the target branch is master or a when running for a tag.
+  echo "Skip script since it is not running on master branch and not running for a tag."
   return 0;
 fi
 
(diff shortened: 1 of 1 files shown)


.ci/update-gh-pages.sh
```diff
@@ -7,18 +7,21 @@ if [ "$TRAVIS" != "true" ]; then
   return 1;
 fi
 
-if [ $TRAVIS_PULL_REQUEST != "false" ]; then
+if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then
```
marcjansen: :+1:

.ci/update-gh-pages.sh
```diff
@@ -7,18 +7,21 @@ if [ "$TRAVIS" != "true" ]; then
...
   # Dont build anything for PR requests, only for merges.
+  echo "Skip script since it processes a Pull Request, not a merge."
```
marcjansen: These will all only be visible if `… > /dev/null 2>&1` is removed from `.travis.yml`, right?

KaiVolland: Yes

chrismayer: Correct. I hope we do not need the logs since everything runs now as expected :wink:

chrismayer:
Thanks for the review, guys! I'll merge now

KaiVolland:
Seems like your fix was not enough. :disappointed:

marcjansen:
I am working on a fix right now.

PR #1148, opened 2019-07-24
Fix ci pipeline

KaiVolland:
This makes the shell script executable.
Compare the git file permission number in the diff.


chrismayer:
Executable seems a good idea for a script we wanna execute... :wink:

marcjansen:
I'll merge to see what happens :crossed_fingers:

marcjansen:
> Seems to work now. At least for "latest".

No sadly it still doesn't. The commit on the `gh_pages` branch was done manually by me, not via Travis.

> Maybe i can fix some bug today, so we can make a new release.

Yeah, do that

marcjansen:
Can we please evaluate whetehr we can use [`gh-pages`](https://www.npmjs.com/package/gh-pages), which we use elsewhere?

KaiVolland:
I'll give it a try.

PR #1152, opened 2019-07-30
Fix i18n for RuleReorderButtons

### LINKED ISSUE #1151: i18n of rule re-ordering buttons does not work
### Description:
It seems like the i18n for the [buttons to re-order the rules](https://github.com/terrestris/geostyler/blob/master/src/Component/RuleTable/RuleReorderButtons/RuleReorderButtons.tsx) in the RuleTable does not work.

Steps to reproduce:

  - Open https://terrestris.github.io/geostyler-demo/
  - Switch to German or Spanisch in the language switcher
  - Hit button `Add Rule` 
  - Hover the non-disabled re-order button 
  - Look at the appearing tooltip

grafik

chrismayer:
Enable i18n for `RuleReorderButtons` component.

Fixes #1151

### DIFF
--- a/src/Component/RuleTable/RuleReorderButtons/RuleReorderButtons.tsx
+++ b/src/Component/RuleTable/RuleReorderButtons/RuleReorderButtons.tsx
@@ -5,6 +5,7 @@ import {
   Rule as GsRule,
 } from 'geostyler-style';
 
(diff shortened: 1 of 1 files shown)


chrismayer:
Thanks for the review @hwbllmnn!

PR #1153, opened 2019-07-30
Add button to clone a rule in RuleTable

### LINKED ISSUE #1150: Add UI tool to clone a rule in RuleTable
### Description:
A tool to clone an existing rule in the `RuleTable` would be a nice improvement. Users can start new rules on existing ones. This can be important in case of having complex symbolizers and/or filters.

Suggestion: Another button `Clone Rule` in the bottom bar of the `RuleTable` UI, which clones all selected rows (by checkbox).

chrismayer:
This adds a tool to clone existing rules in the `RuleTable`, so users can start new rules based on existing ones.

grafik

Closes #1150

### DIFF
--- a/src/Component/Style/Style.tsx
+++ b/src/Component/Style/Style.tsx
@@ -43,6 +43,7 @@ import './Style.css';
 // i18n
 export interface StyleLocale {
   addRuleBtnText: string;
+  cloneRulesBtnText: string;
   removeRulesBtnText: string;
   nameFieldLabel?: string;
   nameFieldPlaceholder?: string;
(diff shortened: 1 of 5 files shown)


src/Component/Style/Style.tsx
```diff
@@ -190,6 +191,35 @@ export class Style extends React.Component<StyleProps, StyleState> {
...
+    } = this.state;
+    const styleClone = _cloneDeep(style);
+
+    // create rules to clone
+    let newRules: GsRule[] = [];
+    styleClone.rules.forEach((rule: GsRule, index: number) => {
+      if (selectedRowKeys.includes(index)) {
+        let ruleClone = _cloneDeep(rule);
+        // TODO We need to ensure that rule names are unique
+        const randomId = Math.floor(Math.random() * 10000);
```
hwbllmnn: Huh ;-) Maybe use https://github.com/kelektiv/node-uuid for this?

marcjansen: Agreed we need to fix this. But in a separate PR, because we had this discussion elsewhere? Do you agree?

jansule: Maybe lodash's [uniqueId()](https://lodash.com/docs/#uniqueId) can be helpful here. We also already use it in this project, as far as I know.

src/locale/es_ES.ts
```diff
@@ -23,6 +23,7 @@ export default {
     },
     GsStyle: {
         addRuleBtnText: 'Añadir regla',
+        cloneRulesBtnText: 'Duplicadas reglas',
```
jansule: Very minor: I think it should be `Duplicar reglas`.

chrismayer: OK, my Spanish is a bit rusty and this is what an online translator gave me :laughing:

chrismayer:
Thanks for you feedback guys! 

The random ID generation was taken from elsewhere in the code, so at least it is harmonized with other code places. I totally agree, that we should introduce something more robust in a separate PR. We should be aware, that there are several places in the GeoStyler code where this has to be changed.

Once Travis is green I'll merge this.

PR #1155, opened 2019-08-01
Update secure variable for gh-pages update

dnlkoch:
This tries to (finally) fix the `gh-pages` auto-update by updating the `GH_TOKEN` variable.


dnlkoch:
:crossed_fingers:

PR #1168, opened 2019-08-16
Introduces manual '.*less' transformation for dist

KaiVolland:
This fixes the build by introducing a manual transformation of the `*.less` files.

As typescript does not handle less files a script is needed to do the transformation itself.

### DIFF
--- a/transform-less.js
+++ b/transform-less.js
@@ -1,0 +1,37 @@
+const fs = require('fs');
+const readdirp = require('readdirp');
+const childProcess = require("child_process");
+
+const replace = require('replace-in-file');
+
+// Transform less
+(async function() {
+  const files = await readdirp.promise('dist', { fileFilter: '*.less' });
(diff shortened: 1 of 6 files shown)


transform-less.js
```diff
@@ -0,0 +1,37 @@
...
+const readdirp = require('readdirp');
+const childProcess = require("child_process");
+
+const replace = require('replace-in-file');
+
+// Transform less
+(async function() {
+  const files = await readdirp.promise('dist', { fileFilter: '*.less' });
+  files.forEach(file => {
+    const out = file.fullPath.replace('less', 'css');
```
dnlkoch: It could be worthwile to check for replacing the file extension only as a file named `CarelessComponent.less` would be renamed to `CarecssComponent.less`.

KaiVolland:
Restored the webpack config and slightly adapted the replace of `less` to `.less`.

PR #1184, opened 2019-09-09
Styleguide - Add seperate page for each component 

jansule:
Creates a separate page for each component. This should enhance the readability of the documentation and should motivate writing more detailed documentation and multiple examples of the same component (where appropriate).


marcjansen:
Will this also decrease the size of the initial pageload for a single component? It used to be very huge at some point in the past.

PR #1185, opened 2019-09-09
Styleguide - Only display non-inherited props

jansule:
Our documentation should only display our own specified props, not those inherited from antd, etc, since these bloat up the documentation quite heavily and make it difficult to identify the important props.

This PR changes that so only our props will be displayed in the documentation.

Example:

`OpacityField` before (goes on for around 50+ props):
image

`OpacityField` after:
image

Code taken from: https://github.com/styleguidist/react-docgen-typescript#parseroptions

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -6,7 +6,12 @@ module.exports = {
   styleguideDir: './build/styleguide',
   assetsDir: './docs',
   propsParser: require('react-docgen-typescript')
-    .withCustomConfig('./tsconfig.json')
+    .withCustomConfig('./tsconfig.json', {propFilter: (prop) => {
+      if (prop.parent) {
+        return !prop.parent.fileName.includes('node_modules');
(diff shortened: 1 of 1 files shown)


jansule:
> Would be cool if there would be an optional checkbox to show/hide inherited docs.

I will take a look at the styleguidist docs and check if there is something like that. If I find something, I will add it in a new PR.

jansule:
Took a look at the react-styleguidist docs and it seems as there is no option to expand/collapse properties. The only way to achieve this would be to write our own styleguidist renderer components, but I don't think this is worth the effort.

marcjansen:
Can we have a link to the component from which we inherit? Or do we habe this already? If so, ignore.

jansule:
We do not have this yet and would have to do this for every component manually. Currently, (property) inheritance can only be checked in the source code.

Issue #650, opened 2018-11-29
Overhaul GeoStyler


KaiVolland:
@jansule and @KaiVolland talked about an overhaul of the component concept.

Currently the user of a component is stuck to its child components, or has to pass properties via an ObjectProperty like `sldRendererProps` in the `Style` component. This feels a bit hackish and doesn't fit the approach of React.

An alternative approach would be to work with renderers:

```jsx
<Style
  style={style}
  nameRenderer={<SpecialStyleNamefield  onChange={/*TODO*/} />}
  rulesRenderer={
    <RuleTable
      onChange={/*TODO*/} 
      ruleRenderer={
        <Rule
          onChange={/*TODO*/} 
          nameRenderer={<NameSpecial onChange={/*TODO*/} />}
          symbolizersRenderer={<SpecialSymbolizersRenderer onChange={/*TODO*/} />}
          filterRenderer={undefined} // Use default FilterRenderer
          scaleDenominatorRenderer={false} // Will not be shown
        />
      }
    />
  }
/>
```

### Paradigms:
1. A component would have a renderer for its (main) child components.
2. If you don't specifiy a renderer (or undefined) a default component is used instead.
3. If you provide `false` nothing will be renderer.

### Open Questions:

1. How to handle the changeListeners and its default behaviour?

Please share your thoughts!

KaiVolland:
With #1179 we show an approach to customize the fields of an Editor:

The `fieldConfig` property allows to replace or hide single Fields of an Editor.

It doesn't allow to change the order of the Elements of an Editor, which might be achieved with a children-based approach. The disadvantag of the childre-based approach would that you'd have to add every field even if no change is needed.

jansule:
See also #1186 for another approach to customize fields. There, react's context feature was used.
This allows the customisation for all customisable components from a global point. The big advantage here is that no props have to be propagated through all parent components. So each component checks for itself if it should be customised. This should avoid errors and keep the code clean. Also, if no provider is being used, the component falls back to its default state.

PR #1186, opened 2019-09-09
Overhaul - Context

jansule:
This example uses react's `context` feature to compose custom editors without the need of rewriting everything again.

The surrounding app would have to wrap a `<Composition.Context value=...>` component around the outermost geostyler component.
value would be an object e.g. like this:

```jsx
 composition: {
        LineEditor: {
          widthField: <CustomWidth />, //replace field with custom component
          colorField: false // deactivate field
        },
        Editor: {
          lineEditor: <CustomLineEditor />, // replace editor with custom component
          markEditor: false, // deactivate editor
        }
      }
```

**Update:** Set up customisation of editors/editor fields following this pattern. Now it is possible to replace or deactivate single fields within a symbolizerEditor. Also, complete editors are replacable. Take a look at `CompositionContext.tsx` to get an idea of all customisable components.

@KaiVolland @chrismayer please review


src/Component/Symbolizer/LineEditor/LineEditor.tsx
```diff
@@ -167,18 +169,44 @@ export class LineEditor extends React.Component<LineEditorProps> {
...
+    const formItemLayout = {
+      labelCol: { span: 8 },
+      wrapperCol: { span: 16 }
+    };
+
+    const path = `LineEditor.${field}Field`;
+    const value = _get(symbolizer, `[${field}]`);
+    const component = CompositionUtil.handleComposition(composition, path, onChange, field, value, fallback);
+
+    if (component == null) {
```
KaiVolland: `===` could be more failsafe

src/Component/CompositionContext/CompositionContext.tsx
```diff
@@ -0,0 +1,17 @@
...
+    dashPatternField?: React.ReactElement | false;
+    dashOffsetField?: React.ReactElement | false;
+    capField?: React.ReactElement | false;
+    JoinField?: React.ReactElement | false;
+    GraphicStrokeField?: React.ReactElement | false;
+    GraphicFillField?: React.ReactElement | false;
+  }
+}
+
+export const CompositionContext = React.createContext({});
```
KaiVolland: missing newline

src/Util/CompositionUtil.ts
```diff
@@ -0,0 +1,37 @@
...
+const _get = require('lodash/get');
+
+class CompositionUtil {
+
+  static handleComposition(composition: Compositions, path: string, onChange: Function, field: string, value: any, fallback: React.ReactElement): React.ReactElement {
+    const compositionValue = _get(composition, path);
+    const globalPath = path.split('.')[path.split('.').length - 1];
+    const globalComposition = _get(composition, globalPath);
+
+    if (compositionValue != undefined) {
```
KaiVolland:
```suggestion
    if (compositionValue !== undefined) {
```

src/Util/CompositionUtil.ts
```diff
@@ -0,0 +1,37 @@
...
+      if (compositionValue != false) {
```
KaiVolland:
```suggestion
      if (compositionValue !== false) {
```

src/Component/Symbolizer/LineEditor/LineEditor.tsx
```diff
@@ -167,18 +169,44 @@ export class LineEditor extends React.Component<LineEditorProps> {
     }
   }
 
+  handleComposition = (composition: Compositions, field:string, onChange: Function, fallback: React.ReactElement) => {
```
KaiVolland: I would be prefer `defaultElement` over `fallback`.

src/Component/Symbolizer/LineEditor/LineEditor.tsx
```diff
@@ -194,93 +222,58 @@ export class LineEditor extends React.Component<LineEditorProps> {
...
-          </Panel>
-        </Collapse>
-      </div>
-    );
+          <CompositionContext.Consumer>
+            {(composition: Compositions) => (
+                <div className="gs-line-symbolizer-editor" >
+                  <Collapse bordered={false} defaultActiveKey={['1']} onChange={(key: string) => (null)}>
+                    <Panel header="General" key="1">
+                      {this.handleComposition(composition, 'color', this.onColorChange, <ColorField />)}
```
KaiVolland: I think it woul be nice to extract these lines to `const` above the `return` to keep the render tree as clean as possible.

KaiVolland:
I think this is a good approach. The components get more complex but this is not avoidable.

If we document this well in the API-docs it should be very easy for users top make use of the customization.

Is this somewhere API breaking or would it be possible to merge this and proceed with further configurations on other components later?

jansule:
> I think this is a good approach. The components get more complex but this is not avoidable.

I think so too. Complexity does not even increase too much, as we handle the main decision making in the utility class.

> If we document this well in the API-docs it should be very easy for users to make use of the customization.

This definitely needs to be done, as we are working with an anti-pattern here. And also users should know about this nice feature.

 
> Is this somewhere API breaking or would it be possible to merge this and proceed with further configurations on other components later?

It is NOT breaking the API, since we provide a default context. Everything stays the same if no context provider was set.

src/Component/Symbolizer/LineEditor/LineEditor.tsx
jansule: Currently I do not think this is possible. The `CompositionContext.Consumer` component injects the context to its children. Right now, I do not see another way of accessing the context inside the render function. The old way of retrieving context (`this.context`) does not return the context, and the new way (`useContext(<Context>)` hook) does not work for classes.

src/Component/CompositionContext/CompositionContext.tsx
jansule: solved in https://github.com/terrestris/geostyler/pull/1186/commits/4fcbbd578e14eb8c7466107a521dd20ed8a860f5

src/Component/Symbolizer/LineEditor/LineEditor.tsx
jansule: solved in https://github.com/terrestris/geostyler/pull/1186/commits/4fcbbd578e14eb8c7466107a521dd20ed8a860f5

src/Component/Symbolizer/Editor/Editor.tsx
```diff
@@ -95,7 +97,7 @@ export class Editor extends React.Component<EditorProps, EditorState> {
     }
   }
 
-  getUiFromSymbolizer = (symbolizer: Symbolizer): React.ReactNode => {
+  getUiFromSymbolizer = (symbolizer: Symbolizer, composition: Compositions): React.ReactNode => {
```
KaiVolland: Please add some Code docs to this and the other newly introduced functions.

src/Util/CompositionUtil.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+        propValue: dummyValue,
+        defaultElement: dummyDefaultElement
+      });
+
+      expect(composition).toBeDefined();
+      expect(composition).toBeFalsy();
+
+    });
+  });
+});
```
KaiVolland: newline please

src/Util/CompositionUtil.ts
```diff
@@ -0,0 +1,86 @@
...
+
+interface CompositionUtilInjectOptions {
+  component: React.ReactElement;
+  onChange: Function;
+  propName: string;
+  propValue: any;
+  onChangeName?: string;
+}
+
+class CompositionUtil {
```
KaiVolland: Please add code docs.

src/Component/CompositionContext/CompositionContext.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
```
KaiVolland: Please add some code docs what this file is about.

src/Component/Symbolizer/Editor/Editor.tsx
jansule: solved in https://github.com/geostyler/geostyler/pull/1186/commits/587b9b93e34174a44cce9a8464bcb50512687670

src/Util/CompositionUtil.spec.tsx
jansule: solved in https://github.com/geostyler/geostyler/pull/1186/commits/587b9b93e34174a44cce9a8464bcb50512687670

src/Util/CompositionUtil.ts
jansule: solved in https://github.com/geostyler/geostyler/pull/1186/commits/587b9b93e34174a44cce9a8464bcb50512687670

src/Component/CompositionContext/CompositionContext.tsx
jansule: solved in https://github.com/geostyler/geostyler/pull/1186/commits/587b9b93e34174a44cce9a8464bcb50512687670

jansule:
@KaiVolland added docs, please review again.

KaiVolland:
Nice job! Please merge at will.

jansule:
closes #650

PR #1221, opened 2019-12-19
Adds Shapefile support to DataLoader

KaiVolland:
This adds handling of Shapefiles to the `DataLoader`.


src/Component/DataInput/DataLoader/DataLoader.tsx
```diff
@@ -102,6 +102,28 @@ export class DataLoader extends React.Component<DataLoaderProps, DataLoaderState
     };
   }
 
+  parseShapefileUploadData = (uploadObject: any) => {
```
dnlkoch: Do we have any type info for the `uploadObject`?

KaiVolland: I added the missingt typings.

PR #1271, opened 2020-05-11
Remove greenkeeper badge

marcjansen:
The title says it all; we no longer (can) use Greenkeeper, and the badge is broken anyhow.

A follow-up would be to have a dependabot badge there instead.


marcjansen:
Thanks for the review.

PR #1332, opened 2021-02-10
Replaces deprecated tslint with eslint

KaiVolland:
## Description

This replaces the deprecated `tslint` package with `eslint` and its corresponding subpackages and configs.

It includes an upgrade of `typescript` to version 4.

## Pull request type

- [x] Dependency updates
- [x] Code style update (formatting, renaming)
- [x] Refactoring (no functional changes, no api changes)
- [x] Build related changes

- [x] No


hwbllmnn (approved):
Nice, please merge!

ahennr (approved):

.eslintignore
```diff
@@ -0,0 +1,8 @@
+# don't ever lint node_modules
+node_modules
+# don't lint build output (make sure it's set to your correct build folder name)
+dist
```
jansule, line 4: We should also add the `browser` folder to this list

src/Util/FilterUtil.spec.ts
```diff
@@ -41,7 +41,7 @@ describe('FilterUtil', () => {
   describe('featureMatchesFilter', () => {
     it('returns true if a feature matches the filter', () => {
       const dummyData = TestUtil.getComplexGsDummyData();
-      let matchingFeature = dummyData.exampleFeatures.features[0];
+      const matchingFeature = dummyData.exampleFeatures.features[0];
```
jansule, line 44: Just out of curiosity: Shouldn't `const` be ok here, as we keep the reference to the same object?

jansule (commented):
Very nice work @KaiVolland! Just a minor remarks. Feel free to merge afterwards

src/Util/FilterUtil.spec.ts
KaiVolland: `const` is not only ok but recommended by eslint. Thats why i switched :)

PR #1345, opened 2021-02-23
Fix Editing in ComparisonFilter

JakobMiksch:
## Description

The input fields of the  `ComparisonFilter` have a bug. It is not possible to make edits in the middle or the beginning of the text. After one character is typed, the cursor instantly jumps to the end of the text. See GIF below. The idea was to type "123" in the middle of the text, but the cursor jumps to end after typing the character "1".

Peek 2021-02-23 12-00

I guess the reason for the bug is that the app creates completely new values of the changed input fields. This happens because of the use of `_cloneDeep()`. This causes the app to forget the previous position of the cursor. This Pull Request removes `_cloneDeep()` and re-creates the component state with changed values (not new ones). Afterwords the input works like expected.

Please list issues or pull requests that the changes you propose are related to. It does not matter if they are still open and/or unmerged, any link is appreciated.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
+++ b/src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
@@ -368,7 +368,7 @@ export class ComparisonFilter extends React.Component<ComparisonFilterProps, Com
       validators
     } = this.props;
 
-    let filter: GsComparisonFilter = _cloneDeep(this.state.filter);
+    let filter: GsComparisonFilter = this.state.filter;
     filter[1] = newAttrName;
 
     const valueFieldVis = this.getValueFieldVis(newAttrName);
@@ -450,7 +450,7 @@ export class ComparisonFilter extends React.Component<ComparisonFilterProps, Com
     const {
       onFilterChange
     } = this.props;
-    let filter: GsComparisonFilter = _cloneDeep(this.state.filter);
+    let filter: GsComparisonFilter = this.state.filter;
     filter[2] = newValue;
 
     // validate value fields


src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
```diff
@@ -450,7 +450,7 @@ export class ComparisonFilter extends React.Component<ComparisonFilterProps, Com
     const {
       onFilterChange
     } = this.props;
-    let filter: GsComparisonFilter = _cloneDeep(this.state.filter);
+    let filter: GsComparisonFilter = this.state.filter;
     filter[2] = newValue;
```
KaiVolland, line 454: Settings the newValue (line 454) will manipulate the JS object in place. This might lead to problems when rerendering the component and comparing its previous value with the new one set via `setState`.

Maybe we can get rid of setting the value but calling setState with the correct value. Can you try to get it working with line [467](https://github.com/geostyler/geostyler/pull/1345/files#diff-98cfd035aec1fee5bb85b5e61c3949d86365efe3e6e09484cbe6a62363724f1fL467) changed to something like this:

```js
filter: [filter[0], filter[1], newValue]
```

Untested code. :smile:

KaiVolland (commented):
Nice enhancement. Thanks for contributing.

`_cloneDeep` is not nice anyway. Please have a look at my comment.

src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
JakobMiksch: I did it like this:
```javascript

// let filter: GsComparisonFilter = this.state.filter;
// filter[1] = newAttrName;

const filter: GsComparisonFilter = [
    this.state.filter[0], 
    newAttrName, 
    this.state.filter[2]
  ];
```

@KaiVolland is that what you meant?

Unfortunately it does not fix the bug.

KaiVolland: Yes that is what i mean but i think it should be:

```js
// let filter: GsComparisonFilter = this.state.filter;
// filter[1] = newAttrName;

const filter: GsComparisonFilter = [
    this.state.filter[0], 
    this.state.filter[1]
    newValue
  ];
```

If this does not work we can still merge your base approach.

JakobMiksch: A "raw" example of a ComparisonFilter is:

```javascript
      filter: [
        "==",
        "my_attribute",
        "my_value"
      ]
```

The "attribute" property has to be at the second position. See also the definition in the [geostyler-style repo](https://github.com/geostyler/geostyler-style/blob/master/index.d.ts#L62-L71).

chrismayer: Also tested the approach suggested by @KaiVolland but unfortunately it still leads to the error with the cursor shift.

chrismayer:
Tested this and it works for me :+1:

src/Component/Filter/ComparisonFilter/ComparisonFilter.tsx
KaiVolland: > A "raw" example of a ComparisonFilter is:
> 
> ```js
>       filter: [
>         "==",
>         "my_attribute",
>         "my_value"
>       ]
> ```
> 
> The "attribute" property has to be at the second position. See also the definition in the [geostyler-style repo](https://github.com/geostyler/geostyler-style/blob/master/index.d.ts#L62-L71).

Strange, as the previous code sets the value at index 2. But if it works… :) I'll have a look at this later.

KaiVolland (approved):

jansule (approved):

Issue #1346, opened 2021-02-24
Default value for InputNumber components


initialdmg:
# Feature Request

**Is your feature request related to a problem? Please describe.**
Add defaultValue prop to InputNumbers. This may cause parse issue, i dont know..., but i think that might make ui clearer with a specific number in  the component 😄 

**Describe the solution you'd like**
Set a specific defaultValue prop on related components

**Describe alternatives you've considered**

**Additional context**

KaiVolland:
Hey @initialdmg this is now possible with the `DefaultValueContext` introduced in version 5.1.0

PR #1374, opened 2021-03-08
Add default value context

hwbllmnn:
## Description

With this PR the default value of 1 is set for opacity values in all symbolizers. This leads to the opacity to be shown in the UI so as not to confuse users. Also the opacity and stroke-opacity fields are added to the fill editor.

## Pull request type

- [ ] Bugfix
- [x] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [ ] I have added or updated tests and documentation, and the test suite passes (run `npm test` locally)

### DIFF
--- a/src/Component/Symbolizer/FillEditor/FillEditor.tsx
+++ b/src/Component/Symbolizer/FillEditor/FillEditor.tsx
@@ -64,6 +64,8 @@ export interface FillEditorLocale {
   outlineWidthLabel?: string;
   graphicFillTypeLabel?: string;
   outlineDasharrayLabel?: string;
+  opacityLabel?: string;
+  strokeOpacityLabel?: string;
 }
 
 interface FillEditorDefaultProps {
@@ -106,6 +108,29 @@ export class FillEditor extends React.Component<FillEditorProps> {
     } = this.props;
     const symbolizer: FillSymbolizer = _cloneDeep(this.props.symbolizer);
     symbolizer.fillOpacity = value;
+    window.console.log('changed:', value);
+    if (onSymbolizerChange) {
+      onSymbolizerChange(symbolizer);
+    }
+  };
+
+  onOpacityChange = (value: number) => {
+    const {
+      onSymbolizerChange
+    } = this.props;
+    const symbolizer: FillSymbolizer = _cloneDeep(this.props.symbolizer);
+    symbolizer.opacity = value;
+    if (onSymbolizerChange) {
+      onSymbolizerChange(symbolizer);
+    }
+  };
+
+  onStrokeOpacityChange = (value: number) => {
+    const {
+      onSymbolizerChange
+    } = this.props;
+    const symbolizer: FillSymbolizer = _cloneDeep(this.props.symbolizer);
+    symbolizer.outlineOpacity = value;
     if (onSymbolizerChange) {
       onSymbolizerChange(symbolizer);
     }
@@ -185,7 +210,9 @@ export class FillEditor extends React.Component<FillEditorProps> {
       outlineColor,
       graphicFill,
       outlineWidth,
-      outlineDasharray
+      outlineDasharray,
+      opacity,
+      outlineOpacity
     } = symbolizer;
 
     const {
@@ -218,8 +245,34 @@ export class FillEditor extends React.Component<FillEditorProps> {
                       composition,
                       path: 'FillEditor.fillOpacityField',
                       onChange: this.onFillOpacityChange,
-                      propName: 'fillOpacity',
+                      propName: 'opacity',
                       propValue: fillOpacity,
+                      defaultElement: <OpacityField />
+                    })
+                  )
+                }
+                {
+                  this.wrapFormItem(
+                    locale.opacityLabel,
+                    CompositionUtil.handleComposition({
+                      composition,
+                      path: 'FillEditor.opacityField',
+                      onChange: this.onOpacityChange,
+                      propName: 'opacity',
+                      propValue: opacity,
+                      defaultElement: <OpacityField />
+                    })
+                  )
+                }
+                {
+                  this.wrapFormItem(
+                    locale.strokeOpacityLabel,
+                    CompositionUtil.handleComposition({
+                      composition,
+                      path: 'FillEditor.strokeOpacityField',
+                      onChange: this.onStrokeOpacityChange,
+                      propName: 'opacity',
+                      propValue: outlineOpacity,
                       defaultElement: <OpacityField />
                     })
                   )
--- a/src/Component/Symbolizer/FillEditor/FillEditor.tsx
+++ b/src/Component/Symbolizer/FillEditor/FillEditor.tsx
@@ -53,6 +53,8 @@ import en_US from '../../../locale/en_US';
 import LineDashField from '../Field/LineDashField/LineDashField';
 import { CompositionContext, Compositions } from '../../CompositionContext/CompositionContext';
 import CompositionUtil from '../../../Util/CompositionUtil';
+import withDefaultsContext from '../../../hoc/withDefaultsContext';
+import { DefaultValues } from '../../DefaultValueContext/DefaultValueContext';
 
 const Panel = Collapse.Panel;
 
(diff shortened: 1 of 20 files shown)


src/Component/Symbolizer/FillEditor/FillEditor.tsx
```diff
@@ -106,6 +108,29 @@ export class FillEditor extends React.Component<FillEditorProps> {
...
+      onSymbolizerChange
+    } = this.props;
+    const symbolizer: FillSymbolizer = _cloneDeep(this.props.symbolizer);
+    symbolizer.opacity = value;
+    if (onSymbolizerChange) {
+      onSymbolizerChange(symbolizer);
+    }
+  };
+
+  onStrokeOpacityChange = (value: number) => {
```
KaiVolland, line 128: Please rename everything with `stroke` to `outline` to be consistent with the geostyler-style.

KaiVolland (changes requested):
Doesn't this lead to unnecessary fields in the target style?
The fields `opacity`, `fill-opacity` and `outline-opacity` are generally optional and would therefore not have to be set.

Please also have a look at my changerequest.

jansule:
> Doesn't this lead to unnecessary fields in the target style?
> The fields `opacity`, `fill-opacity` and `outline-opacity` are generally optional and would therefore not have to be set.

Maybe we should make use of placeholders here. 

This could lead to confusion where default values between different formats differ. But I doubt that this could happen for the opacity field in particular.

src/Component/Symbolizer/FillEditor/FillEditor.tsx
hwbllmnn: Done in dae52fc

hwbllmnn:
> Doesn't this lead to unnecessary fields in the target style?
> The fields `opacity`, `fill-opacity` and `outline-opacity` are generally optional and would therefore not have to be set.

That's true, but I guess the writers could decide whether to write out default values or not.

KaiVolland:
> > Doesn't this lead to unnecessary fields in the target style?
> > The fields `opacity`, `fill-opacity` and `outline-opacity` are generally optional and would therefore not have to be set.
> 
> That's true, but I guess the writers could decide whether to write out default values or not.

But i'm afraid you couldn't determine if a user explicitly wants to have `opacity` set to 1 or not. :thinking: ...

Also if you add default values to the opacity inputfields it would be inconsistent with all other fields. The styles would blow up completly and would contain much unnecessary information.

I think we have to think about this as #1346 also asks for this functionality, but i'm against configuring this directyl in the geostyler.

There should be something like a `DefaultValueContext` where you can get the values from (thx @jansule).

jansule:
@hwbllmnn it would be nice if you could wrap the new context within the `Style` component and add a property `defaultValues` to the component.

So users could define default values as this:

```

const defaultValues = {
  LineSymbolizer: {
    opacity: 0.5
  }
};

<Style
  defaultValues={defaultValues}
/>
```

hwbllmnn:
> @hwbllmnn it would be nice if you could wrap the new context within the `Style` component and add a property `defaultValues` to the component.
> 
> So users could define default values as this:
> 
> ```
> 
> const defaultValues = {
>   LineSymbolizer: {
>     opacity: 0.5
>   }
> };
> 
> <Style
>   defaultValues={defaultValues}
> />
> ```

Just saw this right now, sorry... I've not wrapped this around `Style`, so users can decide for themselves if they want to use it or not.

@KaiVolland I've implemented this for most of the 'simple' properties (and skipped it for now for more complex properties like `graphicFill`).

src/Component/Symbolizer/LineEditor/LineEditor.tsx
```diff
@@ -242,7 +246,7 @@ export class LineEditor extends React.Component<LineEditorProps> {
       <CompositionContext.Consumer>
         {(composition: Compositions) => (
           <div className="gs-line-symbolizer-editor" >
-            <Collapse bordered={false} defaultActiveKey={['1']} onChange={(key: string) => (null)}>
+            <Collapse bordered={false} defaultActiveKey={['1']} onChange={() => (null)}>
```
KaiVolland, line 249: Why is this `onChange` even needed? Did you try removing it?

KaiVolland (approved):
Very nice. Thank you.

src/Component/Symbolizer/LineEditor/LineEditor.tsx
hwbllmnn: You're right, removed it in 1a88ee7

src/Component/Symbolizer/TextEditor/TextEditor.tsx
```diff
@@ -276,6 +280,7 @@ export class TextEditor extends React.Component<TextEditorProps> {
                   onChange: this.onLabelChange,
                   propName: 'value',
                   propValue: symbolizer.label || '',
+                  defaultValue: defaultValues?.TextEditor?.defaultTemplate,
```
jansule, line 283: Minor: shouldn't that rather be `defaultLabel` to be consistent with the geostyler-style property?

src/Component/Symbolizer/WellKnownNameEditor/WellKnownNameEditor.tsx
```diff
@@ -221,6 +224,7 @@ export class WellKnownNameEditor extends React.Component<WellKnownNameEditorProp
                   onChange: this.onColorChange,
                   propName: 'color',
                   propValue: color,
+                  defaultValue: defaultValues?.WellKnownNameEditor?.defaultFillColor,
```
jansule, line 227: see comment above

src/Component/Symbolizer/WellKnownNameEditor/WellKnownNameEditor.tsx
```diff
@@ -234,6 +238,7 @@ export class WellKnownNameEditor extends React.Component<WellKnownNameEditorProp
                   onChange: this.onOpacityChange,
                   propName: 'opacity',
                   propValue: opacity,
+                  defaultValue: defaultValues?.WellKnownNameEditor?.defaultFillOpacity,
```
jansule, line 241: see comment above

src/index.ts
```diff
@@ -104,6 +104,8 @@ import RuleGenerator from './Component/RuleGenerator/RuleGenerator';
 import RuleTable from './Component/RuleTable/RuleTable';
 import { localize } from './Component/LocaleWrapper/LocaleWrapper';
 import { CompositionContext } from './Component/CompositionContext/CompositionContext';
+import { DefaultValueContext } from './Component/DefaultValueContext/DefaultValueContext';
+import withDefaultValues from './hoc/withDefaultsContext';
```
jansule, line 108: `withDefaultValues` -> `withDefaultsContext`

src/index.ts
```diff
@@ -196,5 +198,7 @@ export {
   RuleGenerator,
   RuleTable,
   WidthField,
-  CompositionContext
+  CompositionContext,
+  DefaultValueContext,
+  withDefaultValues
```
jansule, line 203: see comment above

jansule (commented):
Nice work @hwbllmnn. Just a few remarks regarding consistency

src/index.ts
hwbllmnn: Nice catch, fixed in 8953dea

src/index.ts
hwbllmnn: Nice catch, fixed in 8953dea

src/Component/Symbolizer/WellKnownNameEditor/WellKnownNameEditor.tsx
hwbllmnn: Fixed in bbf91de

src/Component/Symbolizer/WellKnownNameEditor/WellKnownNameEditor.tsx
hwbllmnn: Fixed in bbf91de

src/Component/Symbolizer/TextEditor/TextEditor.tsx
hwbllmnn: Fixed in bbf91de

PR #1406, opened 2021-04-01
Fix broken comparison filter

chrismayer:
## Description

This fixes the broken comparison filter, so the state reflects the changes in the UI and the dependent components (e.g.) the `CodeEditor` and ensures that the cursor in the attribute- and value-field does not jump to the end while typing. Achieved by doing the following:

This reverts the changeset introduced in #1345, which solved the unwanted cursor jump (see linked PR for details) but also introduced the missing sync of the filter with dependant components e.g. the `CodeEditor` in the GeoStyler-Demo (compare https://github.com/geostyler/geostyler-demo/issues/180).
Additionally an alternative fix for preserving the cursor position and thus allowing to type in the middle of a word in the attribute- and value-field of the comparison filter is now possible. The fix is based on this thread: https://github.com/facebook/react/issues/955.

  - #1345 
  - https://github.com/geostyler/geostyler-demo/issues/180
  - https://github.com/facebook/react/issues/955

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/Component/Filter/AttributeCombo/AttributeCombo.tsx
+++ b/src/Component/Filter/AttributeCombo/AttributeCombo.tsx
@@ -64,13 +64,14 @@ export interface AttributeComboProps extends Partial<AttributeComboDefaultProps>
 
 interface AttributeComboState {
   value: string | undefined;
+  inputSelectionStart: number;
+  inputSelectionEnd: number;
 }
 
 /**
  * Combobox offering the attributes to be filtered on.
  */
 export class AttributeCombo extends React.Component<AttributeComboProps, AttributeComboState> {
-
   public static defaultProps: AttributeComboDefaultProps = {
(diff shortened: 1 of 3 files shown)


KaiVolland (approved):
Still wonder why this cursor handling is still needed. But i think a lot changed in react since we introduced this component and we could refactor this a lot.
But as it fixes the issues this is totally fine for me know.

jansule (approved):
Thanks for the fix @chrismayer 

It might be useful to have a reference to the thread in the code as a comment. This makes it easier to identify the reason for this logic, in the future. But this is clearly non-blocking.

PR #1418, opened 2021-04-15
Updates react to v17

KaiVolland:
## Description

This updates react to v17.

As `enzyme` does not support react 17 (https://github.com/enzymejs/enzyme/issues/2429) some tests are failing.
So this also introduces `react-testing-library`. At least for the failing tests.

Please list issues or pull requests that the changes you propose are related to. It does not matter if they are still open and/or unmerged, any link is appreciated.

## Pull request type

- [x] Dependency updates
- [x] Build related changes

- [x] No

## Checklist


marcjansen:
Donusers need to upgrade their React as well?

KaiVolland:
> Do users need to upgrade their React as well?

Just tested it with the geostlyer-demo (react 16) and it works as expected. I think its generally not a problem to use react components from different versions in your app. Actually the purpose of react 17 is to enhance the work with multiple react verisons: https://reactjs.org/blog/2020/10/20/react-v17.html#no-new-features

jest/setup.js
```diff
@@ -33,3 +33,17 @@ import Enzyme from 'enzyme';
...
 Enzyme.configure({ adapter: new Adapter() });
+
+Object.defineProperty(window, 'matchMedia', {
+  writable: true,
+  value: jest.fn().mockImplementation(query => ({
+    matches: false,
+    media: query,
+    onchange: null,
+    addListener: jest.fn(), // deprecated
+    removeListener: jest.fn(), // deprecated
```
jansule, line 44: should we use these, if they are deprecated?

jansule (approved):
Very nice!

jest/setup.js
KaiVolland: As it is just a mock this might help if some libraries do use theses functions. We ourselves do not use this.

dnlkoch (approved):

PR #1428, opened 2021-04-28
Refactors node scripts

KaiVolland:
## Description

This refactors the node scripts to seperate and speed up CI and release process.

Previously the use of the (stacked) `pre` hooks led to repeated executes of the tests and unneeded building of the styleguide.
The building of the styleguide is very time consuming and the output it is not included in the package. So there is no need to build it in `prepublishOnly`.
The tests and cleanup are executed by the `np` package. So both are also not needed in `prepublishOnly`.

## Pull request type

- [x] Refactoring (no functional changes, no api changes)
- [x] Build related changes

- [x] No

## Checklist


package.json
```diff
@@ -33,20 +33,21 @@
   ],
   "homepage": "https://github.com/geostyler/geostyler",
   "scripts": {
-    "build": "npm run build:dist && npm run build:browser && npm run build:styleguide",
-    "build:dist": "tsc -p ./ && copyfiles \"./src/**/*.less\" dist --up 1 && node transform-less.js",
-    "build:styleguide": "styleguidist build",
-    "build:browser": "webpack --config browser-build.config.js",
+    "build": "npm run build-dist && npm run build-browser && npm run build-styleguide",
```
jansule, line 36: According to your naming convention, shouldn't this rather be `build-dist:build-browser:build-styleguide`?

marcjansen (approved):
Looks good to me. `npm run lint:test:cleanup:build` is quite a mouthful, though. Thanks for this.

package.json
marcjansen: nah, I'd opt for the short name

PR #1456, opened 2021-05-27
Fix undefined error caused by clearing font list

initialdmg:
## Description

When clear seleted fonts, an undefined error throws by 

https://github.com/geostyler/geostyler-sld-parser/blob/b07d6e71553ca2c99a15a60e7d1bf63df091cefd/src/SldStyleParser.ts#L1509

beacuse `textSymbolizer.font` was assigned with `undefined` .

`font` is not required in the style defination: 

https://github.com/geostyler/geostyler-style/blob/e11fd032b7ee21affd6548acce3ff0568439e185/style.ts#L246

so this property shoud not exist to avoid the error.

## Pull request type

- [x] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/src/Component/Symbolizer/TextEditor/TextEditor.tsx
+++ b/src/Component/Symbolizer/TextEditor/TextEditor.tsx
@@ -134,7 +134,11 @@ export class TextEditor extends React.Component<TextEditorProps> {
       onSymbolizerChange
     } = this.props;
     const symbolizer = _cloneDeep(this.props.symbolizer);
-    symbolizer.font = value.length > 0 ? value : undefined;
+    if (value.length) {
+      symbolizer.font = value;
+    } else {
+      delete symbolizer.font;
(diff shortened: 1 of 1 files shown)


KaiVolland:
Thanks for the contribution.

Actually this seems to be a bug in the SLDSymbolizer and should be fixed there.
Deleting the property or setting it to `undefined` is no big difference i guess, both should not crash the SLDParser.

A nullcheck added to the line mentioned by you should do the trick.

initialdmg:
Yes. And we should also check `textSymbolizer.font[0]` just in case of an empty array is passed in the parser, which might cause a xml2js error(undefined in the node) in my case. Delete this properties is a safer solution.

PR #1543, opened 2022-02-10
Drop the Legacy/CodeEditor

KaiVolland:
### BREAKING CHANGE :exclamation: 

This drops the old and unused `/Legacy/CodeEditor`. I guess it is a breaking change though.


jansule (approved):
Thanks. Yes, it's a breaking change, but cleaning up is always a good thing.

PR #1549, opened 2022-02-11
Update jest to v27

KaiVolland:
This updates jest to version 27.
It also removes `canvas` and `jest-canvas-mock` as it is no longer needed.


jansule (approved):
Very nice!

PR #1585, opened 2022-03-01
Update package-lock version to v2

KaiVolland:
This updaes `package-lock.json` to version 2.
It is recommended to develop geostyler with node version >=16 from now on.


jansule (approved):
Very nice!

Issue #1524, opened 2021-11-19
GeoStyler not working with Node 16


jansule:
# Bug

**Describe the bug**

Currently, it is not possible to install GeoStyler with Node 16.

**To Reproduce**

1. Create an empty directory and `cd` into that directory
2. Run `npm i geostyler`
3. See how it fails with following error message

```
npm ERR! code ERESOLVE
npm ERR! ERESOLVE unable to resolve dependency tree
npm ERR! 
npm ERR! While resolving: undefined@undefined
npm ERR! Found: react@17.0.2
npm ERR! node_modules/react
npm ERR!   peer react@">=16.0.0" from @ant-design/icons@4.7.0
npm ERR!   node_modules/@ant-design/icons
npm ERR!     peer @ant-design/icons@"4.x" from geostyler@7.2.1
npm ERR!     node_modules/geostyler
npm ERR!       geostyler@"*" from the root project
npm ERR!     @ant-design/icons@"^4.6.3" from antd@4.16.13
npm ERR!     node_modules/antd
npm ERR!       peer antd@"4.x" from geostyler@7.2.1
npm ERR!       node_modules/geostyler
npm ERR!         geostyler@"*" from the root project
npm ERR!   peer react@"17.0.2" from react-dom@17.0.2
npm ERR!   node_modules/react-dom
npm ERR!     peer react-dom@">=16.0.0" from @ant-design/icons@4.7.0
npm ERR!     node_modules/@ant-design/icons
npm ERR!       peer @ant-design/icons@"4.x" from geostyler@7.2.1
npm ERR!       node_modules/geostyler
npm ERR!         geostyler@"*" from the root project
npm ERR!       1 more (antd)
npm ERR!     peer react-dom@">=16.9.0" from antd@4.16.13
npm ERR!     node_modules/antd
npm ERR!       peer antd@"4.x" from geostyler@7.2.1
npm ERR!       node_modules/geostyler
npm ERR!         geostyler@"*" from the root project
npm ERR!   1 more (antd)
npm ERR! 
npm ERR! Could not resolve dependency:
npm ERR! peer react@"16.x" from geostyler@7.2.1
npm ERR! node_modules/geostyler
npm ERR!   geostyler@"*" from the root project
npm ERR! 
npm ERR! Fix the upstream dependency conflict, or retry
npm ERR! this command with --force, or --legacy-peer-deps
npm ERR! to accept an incorrect (and potentially broken) dependency resolution.

```

**Expected behavior**

[...]

PR #1594, opened 2022-03-04
Update react, react-dom to v17

KaiVolland:
Updates `react` and `react-dom` to version 17 in the devDependencies.
It also widens the peerDependency range to allow version 16 or later.

This addresses #1524

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,9 +57,9 @@
     "@types/file-saver": "^2.0.1",
     "@types/geojson": "^7946.0.7",
     "@types/lodash": "^4.14.178",
-    "@types/react": "^17.0.3",
+    "@types/react": "^17.0.39",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -141,7 +141,7 @@
     "@ant-design/icons": "4.x",
     "antd": "4.x",
     "ol": "6.x",
-    "react": "16.x",
-    "react-dom": "16.x"
+    "react": ">=16.x",
+    "react-dom": ">=16.x"
```
dnlkoch, line 145: Do we need to set version 17 as required dependency?

dnlkoch (approved):

package.json
KaiVolland: Nope. The goal is, that it runs with react 16 **and** 17.

PR #1687, opened 2022-05-17
Update antd to v4.20.5

KaiVolland:
This updates antd to the latest v4.20.5.

It also adds a workaround to a test which is needed due to this bug in antd:
https://github.com/ant-design/ant-design/issues/35600

Actually this is a good example why testing input elements via className is a bad idea. Maybe adding a test-id to the Element would be a better idea. :)

closes #1662


jansule (approved):
Very nice!

Issue #1375, opened 2021-03-08
Styleguide does not start


JakobMiksch:
# Bug

**Describe the bug**
The styleguide cannot be started using `npm run styleguide`

**To Reproduce**
1. Checkout 9f280731658a636b7252abdb54478418261f9551 (the latest master commit at the time of writing)
2. Run `npm install`
3. Run `npm run styleguide`
4. The styleguide does not show up on `localhost:6060` (the website loads forever)

**Expected behavior**
The styleguide should show up on `localhost:6060`

**Screenshots**

**Desktop**
 - OS: Linux Mint 19 Tara
 - Browser: 
   - Mozilla Firefox 86.0
   - Chromium 87.0
  
**Additional context**

Below some relevant parts of the terminal output. Note the `ℹ ｢wds｣: webpack output is served from undefined`. 

```
[...]
Starting type checking service...                                                              
Using 1 worker with 2048MB memory limit                                                        
Watching: /home/jakob/tmp/geostyle_suite/geostyler/src                                         
ℹ ｢wds｣: Project is running at http://localhost:6060/                                          
ℹ ｢wds｣: webpack output is served from undefined                                               
ℹ ｢wds｣: Content not from webpack is served from /home/jakob/tmp/geostyle_suite/geostyler/docs 
You can now view your style guide in the browser:                                              
                                                                                               
  Local:            http://localhost:6060/                                                     
  On your network:  http://192.168.0.105:6060/                                                 
                                                                                               
[...]
```

JakobMiksch:
**UPDATE**
Around 15 minutes after starting `npm run styleguide` I noticed that the styleguide is finally running. (I actually discovered it by accident ... ) And the terminal output shows a ` DONE  Compiled successfully! `, I could not see that before.

I assume the command `npm run styleguide` takes very long. 
Is this normal or is there anything I am missing?

KaiVolland:
Thanks for the hint. I wouldn't call it "normal". We should have a look at this.

KaiVolland:
Hey @JakobMiksch,
we upgraded `react-stylguidist` in https://github.com/geostyler/geostyler/pull/1393. Would be nice if you could check if it runs on your setup, too now.

JakobMiksch:
I ran it again on your branch. It works, but takes ~10 minutes to start.

chrismayer:
[...]

PR #1739, opened 2022-06-09
Speed up start of the styleguidist

dnlkoch:
## Description

After fiddling around with the current configuration, it looks like the `react-docgen-typescript` is slowing down the initial build of the styleguidist tremendously while checking the props/types of any parent component (e.g. the extended antd types).

This suggests to skip this check in `development` mode which will result in decent startup times (before: ~180s. after: ~20 s), **BUT** the list of props might not be complete. There seems to be a bug while detecting the props of function components as well, see [here](https://github.com/styleguidist/react-styleguidist/issues/1989).

#1375.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [x] Build related changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -28,23 +28,27 @@
 
 const path = require('path');
 const webpackConfig = require('./webpack.dev.config');
+const rdt = require('react-docgen-typescript');
 
 module.exports = {
   title: 'GeoStyler',
   styleguideDir: './build/styleguide',
   webpackConfig: {
     ...webpackConfig,
-    mode: 'production'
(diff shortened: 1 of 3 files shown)


KaiVolland (approved):
Two more big issue with styleguidist detected by your work. We should really investigate if we can replace it with something similar.

The PR is a good option anyway as it just changes the dev setup if im right. Feel free to merge.

dnlkoch:
Yes, I agree, we should definitely check for an alternative.

Issue #1769, opened 2022-06-16
geostyler-demo doesn't start or build on windows since changes in webpack configuration


slafayIGN:
# Bug

Since recent changes in the way `geostyler-demo` is built, we have troubles making it start. (working on last commit of master branch [91bd6b05](https://github.com/geostyler/geostyler-demo/commit/91bd6b052c637db9dda344a07929de2dd58fe5ae))

* `npm run start` : Webpack takes long minutes building thousands of dependencies and modules in windows environment and never ends. Outputs look like:
```<s> [webpack.Progress] 46% building 2/3 entries 4440/4600 dependencies 426/1374 modules```

* `npm run start`: is ok in a linux environment without `npm link`. It fails when there is a `npm link` in use
* `npm run build` : is not usable in windows environment because it starts with a linux like environment variable definition `NODE_ENV=production`

Also the "troubleshooting" part in [README.md](https://github.com/geostyler/geostyler) is not relevant anymore because there is no `react-scripts` dependency anymore in `geostyler-demo`.

**Desktop:**
 - OS: windows (mostly)
 
 Maybe those webpack changes you are making are still a work in progress. Sorry if this is a bit to soon to submit this issue.

KaiVolland:
You are right. I'll update the README.md and hopefully it will work again on windows.

slafayIGN:
Thank you for the readme update.

Although it doesn't seem to solve this issue on windows.

I think the issue lies in the exclude line in webpack config : https://github.com/geostyler/geostyler-demo/blob/master/config/webpack.common.config.js#L11

jansule:
Thanks for the udpate @slafayIGN. Did you find a workaround for windows?

PR #1772, opened 2022-06-17
updates Developer Guide in README.md

KaiVolland:
## Description

This updates the Developer Guide section of the README.md. 

Some configurations in the `geostyler-demo` have changed due to the removal of react-scripts https://github.com/geostyler/geostyler-demo/pull/326.

@slafayIGN please have a look. Lets see if this works for you on windows, too.

This addresses #1769.

## Pull request type

- [ ] Bugfix
- [ ] Feature
- [ ] Dependency updates
- [ ] Build related changes
- [x] Documentation content changes
- [ ] Other (please describe)

- [ ] Yes
- [x] No

## Checklist

- [x] I understand and agree that the changes in this PR will be licensed under the [BSD 2-Clause License](https://github.com/geostyler/geostyler/blob/main/)
- [x] I have followed the [guidelines for contributing](https://github.com/geostyler/geostyler/blob/main/CONTRIBUTING.md)
- [x] The proposed change fits to the content of the [code of conduct](https://github.com/geostyler/geostyler/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,9 +151,9 @@ npm i
 npm link geostyler
 ```
 
-When working with npm link it may happen that some tools like webpack or typescript don't know how to resolve packages that are used in booth packages (in this case `geostyler` and `geostyler-style`). So we have to configure this in the `geostyler-demo`:
+When working with npm link it may happen that some tools like webpack or typescript don't know how to resolve packages that are used in both packages (in this case `geostyler` and `geostyler-style`). So we have to configure this in the `geostyler-demo`:
 
-Replace `rootDird` with `rootDirs` and add the linked packages to the `rootDirs` in `tsconfig.json`. If you link some parsers too just add them aswell.
+Replace `rootDir` with `rootDirs` and add the linked packages to the `rootDirs` in `tsconfig.json`. If you link some parsers too just add them as well.
 
 ```diff
 --- a/tsconfig.json
@@ -172,7 +172,7 @@ Replace `rootDird` with `rootDirs` and add the linked packages to the `rootDirs`
      "noImplicitReturns": true,
 ```
 
-Then uncomment or add resolve aliases to the wepackconfig:
+Then uncomment or add resolve aliases to the webpack config:
 
 ```diff
 --- a/config/webpack.common.config.js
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -149,8 +149,52 @@ git clone https://github.com/geostyler/geostyler-demo.git
 cd geostyler-demo
 npm i
 npm link geostyler
+```
+
+When working with npm link it may happen that some tools like webpack or typescript don't know how to resolve packages that are used in booth packages (in this case `geostyler` and `geostyler-style`). So we have to configure this in the `geostyler-demo`:
```
jansule, line 154:
```suggestion
When working with npm link it may happen that some tools like webpack or typescript don't know how to resolve packages that are used in both packages (in this case `geostyler` and `geostyler-style`). So we have to configure this in the `geostyler-demo`:
```

README.md
```diff
@@ -149,8 +149,52 @@ git clone https://github.com/geostyler/geostyler-demo.git
...
+     "rootDirs": [
+-      "./src/"
++      "./src/",
++      "../geostyler",
+     ],
+     "forceConsistentCasingInFileNames": true,
+     "noImplicitReturns": true,
+```
+
+Then uncomment or add resolve aliases to the wepackconfig:
```
jansule, line 175:
```suggestion
Then uncomment or add resolve aliases to the webpack config:
```

README.md
```diff
@@ -161,34 +205,9 @@ Changes will automatically be updated in the browser. Please also provide tests
...
-      },
-    [...]
-    },
-[...]
-```
-If there is an issue with a UI component, you may need to do the same for the `antd` module and add the alias:
-```javascript
-'antd': path.join(__dirname, '../../', 'antd'),
-```
+If the demo does not start but shows the above error it means that `geostyler-demo` and `geostyler` are using different react sources. Please make sure to have the react alias in the `webpack.common.config.js` configured correctly.
```
jansule, line 208:
```suggestion
If the demo does not start but shows the above error, it means that `geostyler-demo` and `geostyler` are using different react sources. Please make sure to have the react alias in the `webpack.common.config.js` configured correctly.
```

jansule (changes requested):
Thanks @KaiVolland 

I'm wondering, if we should provide a description on how to work with styleguidist (or its successor) instead of using the demo.

This would probably still require using `npm link` when working with unpublished versions of the parsers.

I will add a dedicated issue for this matter, so this is nothing you have to address in this PR.

jansule (approved):
Very nice!

I also already created the issue: https://github.com/geostyler/geostyler/issues/1790

PR #1962, opened 2022-11-16
Extend version range to allow ol >=6

KaiVolland:
This extends the version range for ol in the peerDependencies to allow the use of ol 7.


jansule (approved):
Very nice!

PR #2062, opened 2023-01-18
Fix translation

FritzHoing:
## Description
- Proposed imports need to be imported as default for the localize wrapper to work as intended.
- Adds missing dynamic translation to the ScaleDenominator files.

Breaking:
- remove label property

- [x] Yes
- [ ] No


src/Component/ScaleDenominator/MaxScaleDenominator.tsx
```diff
@@ -51,8 +58,9 @@ export interface MaxScaleDenominatorProps extends Partial<MaxScaleDenominatorDef
  * Input field for the maximum scale of a rule.
  */
 export const MaxScaleDenominator: React.FC<MaxScaleDenominatorProps> = ({
-  label = 'Max. Scale',
-  placeholder = 'Enter max. Scale (Optional)',
+  placeholder,
+  locale = en_US.ScaleDenominator,
+  label = locale.maxScaleDenominatorLabelText,
```
KaiVolland, line 63: I think the extra `label` prop should be removed here and we should use `locale.maxScaleDenominatorLabelText` in the JSX element instead.

src/Component/ScaleDenominator/MaxScaleDenominator.tsx
```diff
@@ -63,16 +71,19 @@ export const MaxScaleDenominator: React.FC<MaxScaleDenominatorProps> = ({
   }
 
   return (
-    <Form.Item className="gs-max-scaledenominator" label={label} colon={false} {...formItemLayout} >
+    <Form.Item className="gs-max-scaledenominator"
+      label={label}
+      colon={false}
+      {...formItemLayout} >
```
KaiVolland, line 77:
```suggestion
      {...formItemLayout}
  >
```

src/Component/ScaleDenominator/MinScaleDenominator.tsx
```diff
@@ -51,8 +58,9 @@ export interface MinScaleDenominatorProps extends Partial<MinScaleDenominatorDef
  * Input field for the minimum scale of a rule.
  */
 export const MinScaleDenominator: React.FC<MinScaleDenominatorProps> = ({
-  label = 'Min. Scale',
-  placeholder = 'Enter min. Scale (Optional)',
+  placeholder,
+  locale = en_US.ScaleDenominator,
+  label = locale.minScaleDenominatorLabelText,
```
KaiVolland, line 63: see above

src/Component/ScaleDenominator/MinScaleDenominator.tsx
```diff
@@ -63,16 +71,22 @@ export const MinScaleDenominator: React.FC<MinScaleDenominatorProps> = ({
   }
 
   return (
-    <Form.Item className="gs-min-scaledenominator" label={label} colon={false} {...formItemLayout}>
+    <Form.Item
+      className="gs-min-scaledenominator"
+      label={label}
+      colon={false}
+      {...formItemLayout}>
```
KaiVolland, line 78:
```suggestion
      {...formItemLayout}
  >
```

KaiVolland (commented):
Nice.
But as you're refactoring it already we might consider to remove the `label` prop completly as it is part of the locale.
This would leed to a breaking change but we are doing a bigger refactoring with #1998 anyway so that should be no problem.

FritzHoing:
Thank you for the review @KaiVolland !

KaiVolland (approved):

src/Component/ScaleDenominator/MinScaleDenominator.tsx
```diff
@@ -63,16 +68,23 @@ export const MinScaleDenominator: React.FC<MinScaleDenominatorProps> = ({
...
-        placeholder={placeholder}
+        placeholder={locale.minScaleDenominatorPlaceholderText}
         onChange={onChange}
       />
     </Form.Item>
   );
 };
 
-export default MinScaleDenominator;
+// export default MinScaleDenominator;
```
jansule, line 88: Minor: Please remove this line

jansule (changes requested):
Very nice @FritzHoing. Just one remark

src/Component/ScaleDenominator/MinScaleDenominator.tsx
FritzHoing: Good catch! Thank you @jansule

jansule (approved):

PR #2080, opened 2023-02-17
feat: use vertical form layout

jansule:
## Description

Form items now use a vertical layout, which means that the labels are on top of the input field. Before, they were located in the same line as the input fields. This should finally fix all form inconsistencies and simplifies this strongly.

Symbolizer Editor

image

Card Style

image

Rule Generator

image

- [ ] Yes
- [x] No

### DIFF
--- a/src/Component/BulkEditor/BulkEditor.tsx
+++ b/src/Component/BulkEditor/BulkEditor.tsx
@@ -95,16 +95,10 @@ export const BulkEditor: React.FC<BulkEditorProps> = ({
     onStylePropChange('image', newImage);
   };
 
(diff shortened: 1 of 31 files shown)


KaiVolland:
Passing the FormItmeProps has no downside does it? I would not remove this feature. Even more if it is a breaking change.

jansule:
> Passing the FormItmeProps has no downside does it? I would not remove this feature. Even more if it is a breaking change.

Removed the breaking change. Min-/MaxScaleDenominator still extend FormItemProps.

KaiVolland (approved):

PR #2121, opened 2023-05-08
feat: use GeoStylerContext composition on IconEditor

jansule:
## Description

This makes use of the GeoStylerContext.composition in `<IconEditor>`.

BREAKING CHANGE: This removes the support of the deprecated CompositionContext in favor of the new GeoStylerContext for the IconEditor. This also removes the `defaultValues` property from IconEditor. Please use GeoStylerContext.composition instead.

geostyler-iconeditor-context

- [x] Yes
- [ ] No

### DIFF
--- a/src/Component/Symbolizer/Field/OffsetField/OffsetField.tsx
+++ b/src/Component/Symbolizer/Field/OffsetField/OffsetField.tsx
@@ -33,7 +33,7 @@ import { InputNumberProps } from 'antd/lib/input-number';
 import FieldUtil from '../../../../Util/FieldUtil';
 
 // non default props
-export interface OffsetFieldProps extends InputNumberProps {
+export interface OffsetFieldProps extends InputNumberProps<number> {
   offset?: number;
 }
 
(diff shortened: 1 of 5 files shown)


hwbllmnn (approved):

jansule:
> A changelog (or migration notes) would be nice.

This is part of the commit message in the block `BREAKING CHANGE`. 

> BREAKING CHANGE: This removes the support of the deprecated CompositionContext in favor of the new GeoStylerContext  for the IconEditor. This also removes the defaultValues property from IconEditor. Please use GeoStylerContext.composition instead.

Would you prefer more detailed information?

marcjansen:
> Do you prefer more detailed information?

Generally speaking: yes. I am a fan of actual files that contain the changes. If we communicate this clearly in the release notes, then fine for me, but a `CHANGELOG.md` would be my choice. 

In the end you can decide, of course.

jansule:
> > Do you prefer more detailed information?
> 
> Generally speaking: yes. I am a fan of actual files that contain the changes. If we communicate this clearly in the release notes, then fine for me, but a `CHANGELOG.md` would be my choice.
> 
> In the end you can decide, of course.

A changelog will be created automatically on every release. So we do not have to include this in our PRs. See also https://github.com/geostyler/geostyler/blob/main/CHANGELOG.md

> Additionally, I think the commit message should have started with breaking, right?

No, the indicator for breaking changes is a footer starting with `BREAKING CHANGE:`. See https://www.npmjs.com/package/semantic-release

PR #2140, opened 2023-05-10
Replace react-rnd windows with modals

dnlkoch:
## Description

This replaces all `react-rnd` window components with [antd Modals](https://ant.design/components/modal).

Peek 2023-05-10 11-38

Please review.

- [ ] Yes
- [x] No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -38,7 +38,6 @@
         "lodash": "^4.17.21",
         "monaco-editor": "^0.38.0",
         "react-color": "^2.19.3",
-        "react-rnd": "^10.4.1",
         "typescript-json-schema": "^0.56.0"
       },
       "devDependencies": {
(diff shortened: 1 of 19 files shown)


KaiVolland:
Is it possible to stick the Modal to a dom Object? Just like the windows did before? Otherwise yo can't see which symbolizer you are currently editing.

hwbllmnn (approved):

dnlkoch:
> Is it possible to stick the Modal to a dom Object? Just like the windows did before? Otherwise yo can't see which symbolizer you are currently editing.

This would be possible, but I think we would run into the same issues as before (e.g. the positions wouldn't match in integrated environments like the GeoServer extensions).

@jansule and I discussed this and would prefer a separate symbolizer preview inside the editor instead - which should be implemented in an upcoming PR.

PR #2147, opened 2023-05-11
Introduces StringExpressionSelect

KaiVolland:
Introduces the `StringExpressionSelect` which can be used for SelectFields in Editors where `Expression`s are supported.

BREAKING CHANGE:
- transforms the value property for some input fields from there propertyname (`cap`, `join`) to `value`

- [x] Yes
- [ ] No


hwbllmnn (approved):

jansule (changes requested):
Code looks good to me, thanks! Please add the `BREAKING CHANGE` footer to the commit that introduced that breaking change. Please also include migration info there.

KaiVolland:
@jansule I can squash merge this and add the Breaking Change info there.

jansule (approved):

Issue #1805, opened 2022-07-05
Use Issue Forms instead of Templates


jansule:
We can enhance the creation of issues for this repo by providing issue forms. See https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

Our templates should be enhanced anyways, so we might want to switch.

KaiVolland:
This is a feature i still really like. We should do this in the near future.

slafayIGN:
🚀 https://github.com/geostyler/geostyler/issues/new/choose

PR #2150, opened 2023-05-12
chore: replace issue templates with issue forms

slafayIGN:
The issue forms keep the same names, labels and assignees than the previous issue templates.

Related issue #1805

Note the commits here are not using the commit name convention. This PR should be squashed and merged.


jansule (approved):
That's great! Thanks @slafayIGN

Issue #1940, opened 2022-10-29
implement visibility style property to display or hide a specific rule.


iamgeoknight:
# Feature Request
I am currently using geostyler for one of my project. I would like to have **visibility** property implemented in geostyler rules for symbolizers. If **visibility** is true, that symbolizer will be displayed on map. If **visibility** is false, them symbolizer should be hidden.

Something like below:

{
    
    "rules": [               
        {                
            "name": "COASTAL",
            "filter": [
                "==",
                "TYPE",
                "COASTAL"
            ],
            "symbolizers": [
                {
                    "kind": "Line",
                    "color": "#339FFF",
                    "width": 6,
                    "visibility": false
                }                
            ]
        },
        {                
            "name": "MEXICAN",
            "filter": [                
                "==",
                "TYPE",
                "MEXICAN"
            ],
            "symbolizers": [
                {
                    "kind": "Line",
                    "color": "#D4AA2E",
                    "width": 6,
                    "visibility": true
                }
            ]
        },
        {                
            "name": "CANADIAN",
            "filter": [                
                "==",
                "TYPE",
                "CANADIAN"
            ],
            "symbolizers": [
                {
                    "kind": "Line",
                    "color": "#FF5E33",
                    "width": 6,
                    "visibility": true
                }
            ]
        }
    ]
}

jansule:
Thanks @iamgeoknight

+1 from my side.

The geostyler-style already provides that property for symbolizers, so we might only need to add a visibility switch to the UI (and maybe some parsers, where support is still missing). Are you planning on providing a PR for this?

miccoh1994:
@jansule I need this, can I submit a PR? Would be happy to submit change to the openlayers parser repo as well.

jansule:
@miccoh1994 Sure, submitting a PR would be great! Thanks!

KaiVolland:
This is implemented.

PR #2155, opened 2023-05-13
feat: visibility field

miccoh1994:
## Description

Adds the VisibilityField to the Symbolizer Editor to toggle the visibility of each symbolizer.

 #1940 

Since the visibility property is available on each Symbolizer in geostyle I have placed it under the Kind select in the SymbolizerEditor.

- [ ] Yes
- [ ] No
- [x] I am unsure (no worries, we'll find out)


KaiVolland:
Hey @miccoh1994 :wave: and thanks for the contribution. This looks great.

KaiVolland:
@miccoh1994 Very nice. I checked it in the styleguide and just have one minor remark.

I think it would be better that the checkbox is only unchecked if the symbolizer is configured with `visibility: false`. If not configured at all or configured with `undefined` or `true` the checkbox should be checked. Could you adapt this?

miccoh1994:
@KaiVolland thanks that makes sense, I did consider that but wasn't sure. I will work on it this weekend.

KaiVolland:
@miccoh1994 will you be able to adapt this soon? Would be cool to have this feature in the next release.

miccoh1994:
@KaiVolland adapted, not sure why the commit lint fails though.

KaiVolland (approved):

KaiVolland:
Commit lint expects the commit message to start lowercase. I used squash commit to evade this.

PR #2191, opened 2023-05-26
Use Node.js versions 18.x and 20.x for workflows

marcjansen:
This PR updates the versions of Node.js used during CI runs.


KaiVolland:
Thanks.

PR #2213, opened 2023-05-30
Use antd ColorPicker for `ColorField`

### LINKED ISSUE #2163: Ant Design ColorPicker
### Description

Antd now (version 5.5) has its own [ColorPicker](https://ant.design/components/color-picker).
IMHO we should use it in our `ColorField`.

1. It fits the visual appearance.
2. We can reduce our dependencies.

### Example

_No response_

### Alternatives

_No response_

### Additional Context

https://ant.design/components/color-picker

KaiVolland:
## Description

This replaces the ColorPicker from 'react-color' with the antd `ColorPicker`.

It also changes the property name for the color value from 'color' to 'value' which is a breaking change.

This closes #2163 

- [x] Yes
- [ ] No


src/Component/Symbolizer/Field/ColorField/ColorField.tsx
```diff
@@ -26,108 +26,71 @@
...
-    setColorPickerVisible(!colorPickerVisible);
-  };
-
-  const onChangeComplete = (colorResult: ColorResult) => {
-    if (onChange) {
-      onChange(colorResult.hex);
+  const onColorPickerChange = useCallback((_: any, hex: string) => {
+    // contains 0% opacity --> should only happen when clear is clicked
+    if (hex?.length === 9) {
+      onChange?.(undefined);
```
jansule, line 64: Wouldn't it be easier and more readable to provide a default function when destructuring the `onChange` prop?

I.e.

```js
const ColorField = ({
  //...
  onChange = () => {},
  //...
}) => {}
```

src/Component/Symbolizer/Field/ColorField/ColorField.tsx
```diff
@@ -26,108 +26,71 @@
...
-    }
   }
 
+  const btnStyle: React.CSSProperties = {
+    backgroundColor: value || defaultValue,
+    color: textColor
+  };
+
   return (
     <div className="editor-field color-field">
```
jansule, line 83: Not directly related, but AFAIK, our classes usually have a `gs-` prefix. Maybe, this is a good chance to add this here as well.

KaiVolland:
I'll squash merg this and adapt the commit message to the one you suggested.

PR #2217, opened 2023-05-31
Update outline opacity label for en_US

KaiVolland:
## Description

This updates the en_US label for the `outlineOpacity` field. It also updates the label for the switch in the example.

- [ ] Yes
- [x] No


jansule (approved):
Very nice

PR #2270, opened 2023-06-13
Update geostyler dependencies

KaiVolland:
## Description

This updates the geostyler dependencies.

Due to some changes in the `geostyler-style` there are some adjustments needed. (seems like we should have released a major version in geostyler-style).
So it introduces a `FunctionUtil` (more or less a copy of the [OlStyleUtil](https://github.com/geostyler/geostyler-openlayers-parser/blob/master/src/Util/OlStyleUtil.ts)) and adds some throws in the `FilterUtil`.

- [ ] Yes
- [x] No


src/Util/FilterUtil.ts
```diff
@@ -342,6 +357,9 @@ class FilterUtil {
       newFilter = newFilter as CombinationFilter;
       newFilter.push(addedFilter);
     } else {
+      if (!Array.isArray(newFilter)) {
+        throw new Error(`Can not add filter to filter ${rootFilter}. Root filter has to be an array.`);
```
hblitza, line 361:
```suggestion
        throw new Error(`Cannot add filter to filter ${rootFilter}. Root filter has to be an array.`);
```

src/Util/FilterUtil.ts
```diff
@@ -372,6 +393,9 @@ class FilterUtil {
       case 'or':
         if (previousFilter && (previousFilter[0] === '&&' || previousFilter[0] === '||' )) {
           addedFilter = previousFilter;
+          if (!Array.isArray(addedFilter)) {
+            throw new Error('Can not change filter. Filter is not an array.');
```
hblitza, line 397:
```suggestion
            throw new Error('Cannot change filter. Filter is not an array.');
```

hblitza:
LGTM in general.

hblitza (approved):

src/Util/FunctionUtil.ts
```diff
@@ -0,0 +1,254 @@
...
+  T extends GeoStylerUnknownFunction ? any :
+  never;
+
+/**
+ * @class FunctionUtil
+ */
+class FunctionUtil {
+
+  public static evaluateFunction<T extends GeoStylerFunction>(func: T, feature?: Feature): ReturnType<T> {
+    if (func.name === 'property') {
```
ahennr, line 56: Is a null check for `func` needed?

src/Util/FunctionUtil.ts
```diff
@@ -0,0 +1,254 @@
...
+    }
+    const args = func.args.map(arg => {
+      if (isGeoStylerFunction(arg)) {
+        return FunctionUtil.evaluateFunction(arg, feature);
+      }
+      return arg;
+    });
+    switch (func.name) {
+      case 'abs':
+        return Math.abs(args[0] as number);
```
ahennr, line 135: What happens if no number is passed in the arguments? Should the args be checked here, e.g. using `_isFinite(…)`?

KaiVolland: This should not be a problem. This should just return `NaN` when you pass in an illegal `GeoStylerNumberFunction`. Classy shit in shit out. :smile:

PR #2284, opened 2023-06-21
Fix OlRenderer

KaiVolland:
## Description
This fixes a strange behaviour when using a geostyler function in a symbolizer property with the `RuleTable` component. The source of this behaviour was this bug: https://github.com/geostyler/geostyler-openlayers-parser/issues/740

To make sure the parser won't affect the passed style at all i added a `structuredClone` to the `OlRenderer`.

This also includes some minor refactoring.

- [ ] Yes
- [x] No


jansule (approved):
Very nice!

Issue #2290, opened 2023-06-23
Adjust colorPicker width


jansule:
### Description

The colorpicker does not stretch to the full width of the editor, whereas the other input fields do. This is probably introduced when switching to antd colorpicker.

### Expected behaviour

The colorpicker should stretch to the full width of the editor, just like the other input fields do

### How to reproduce

Example from https://geostyler.github.io/geostyler-demo/

![image](https://github.com/geostyler/geostyler/assets/12186477/3985afad-4e1e-4e3b-a8e3-b8644805aca7)

PR #2299, opened 2023-06-26
fix: set ColorPicker to full width

jansule:
## Description

This sets the color picker to full width in order to be consistent with other input fields.

image

Popup remains small:

image

closes https://github.com/geostyler/geostyler/issues/2290

- [ ] Yes
- [x] No


KaiVolland:
Thx.

Issue #2345, opened 2023-08-15
Geostyler sends request to online cdn


vahidbabaveisi:
### Description

Hello,

I contfront with a problem, I developed my react app using geostyler. Everything is ok when we work online, but as we are offline, geostyler app stops working since it sends request to https://cdn.jsdelivr.net/npm/monaco-editor@0.36.1/min/vs/loader.js for loading monaco editor. I can not solve this. Please help me.

### Expected behaviour

I expect that when we build react app, It works stand alone even when we are offline.

### How to reproduce

Web browser-react app

KaiVolland:
Unfortunately this request is done by the monaco editor. But you can try to overwrite this handling: https://github.com/suren-atoyan/monaco-react#use-monaco-editor-as-an-npm-package.

jansule:
I think this is something that needs to be adjusted in GeoStyler itself and not in applications using GeoStyler. We should check, if there is a way to circumvent this. Maybe the link provided is already what we need.

vahidbabaveisi:
Thanks @KaiVolland . It solved my problem. Now, it is fully stand-alone and works without any connection to the internet. You suvived me!

jansule:
unstale

PR #2346, opened 2023-08-18
fix: replace monaco CDN loader with npm package

jansule:
## Description

This makes use of the monaco loader and configs provided via npm. By default, the editor requests those via http requests to a CDN. Now, everything is included in the build, so no more requests will be sent.

See also https://github.com/suren-atoyan/monaco-react#use-monaco-editor-as-an-npm-package.

Fixes https://github.com/geostyler/geostyler/issues/2345

- [ ] Yes
- [x] No

### DIFF
--- a/src/Component/CodeEditor/CodeEditor.tsx
+++ b/src/Component/CodeEditor/CodeEditor.tsx
@@ -33,7 +33,32 @@ import React, {
   useState
 } from 'react';
 
-import Editor, { useMonaco } from '@monaco-editor/react';
+import * as monaco from 'monaco-editor';
+import Editor, { loader } from '@monaco-editor/react';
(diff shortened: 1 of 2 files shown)


src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -33,7 +33,32 @@ import React, {
   useState
 } from 'react';
 
-import Editor, { useMonaco } from '@monaco-editor/react';
+import * as monaco from 'monaco-editor';
```
KaiVolland, line 36:
```suggestion
import monaco from 'monaco-editor';
```
should work too

jansule:
unstale

PR #2387, opened 2023-11-21
feat: add editor visibility field to composition

jansule:
## Description

This moves the visibility field to the editors and adds the visibility fields to the geostyler composition context.

`visibility: false`

image

`visibility: true`

image

- [ ] Yes
- [x] No

### DIFF
--- a/src/Component/Symbolizer/Editor/Editor.tsx
+++ b/src/Component/Symbolizer/Editor/Editor.tsx
@@ -29,7 +29,6 @@
 import React from 'react';
 
 import {
-  Expression,
   Symbolizer,
   SymbolizerKind
 } from 'geostyler-style';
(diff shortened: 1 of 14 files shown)


src/Component/Symbolizer/Editor/Editor.tsx
```diff
@@ -75,6 +79,8 @@ export interface EditorComposableProps {
   rasterEditor?: {
     visibility?: boolean;
   };
+  // TODO support default value for visibilityField
```
KaiVolland, line 82: Do we really need a configurable default value for this? IMHO we don't need this TODO here.

KaiVolland (approved):

src/Component/Symbolizer/Editor/Editor.tsx
jansule: I do not see a clear use-case here, but why restrict this field while the others do have support for this?

KaiVolland (approved):

PR #2388, opened 2023-11-22
fix: set vertical layout in BulkEditor

jansule:
## Description

This wraps the form items in `<BulkEditor>` into a `<Form>` and sets its layout to `'vertical'` so that the form layout will be consistent with all the other forms. It also adjusts the layout of the symbol selection.

image

- [ ] Yes
- [x] No

### DIFF
--- a/src/Component/RuleGenerator/RuleGenerator.tsx
+++ b/src/Component/RuleGenerator/RuleGenerator.tsx
@@ -50,6 +50,7 @@ import {
   useGeoStylerData,
   useGeoStylerLocale
 } from '../../context/GeoStylerContext/GeoStylerContext';
+import { getFormItemConfig } from '../../Util/FormItemUtil';
 
 export type LevelOfMeasurement = 'nominal' | 'ordinal' | 'cardinal';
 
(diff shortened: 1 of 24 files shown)


KaiVolland (commented):
I would avoid using form at all here as the `<form>` is not a layout element in the first place but influences the handling a lot.

Maybe you can have a look at the other editors to achieve the same result without using a form.

jansule:
> I would avoid using form at all here as the `<form>` is not a layout element in the first place but influences the handling a lot.
> 
> Maybe you can have a look at the other editors to achieve the same result without using a form.

The other editors are all wrapped in forms as well. But I agree, the form was mainly added for using the layout property, which we also can achieve via css only.

jansule:
Removed the forms. @KaiVolland could you take another look?

src/Component/RuleGenerator/RuleGenerator.tsx
```diff
@@ -171,146 +172,157 @@ export const RuleGenerator: React.FC<RuleGeneratorProps> = (props) => {
...
+              ? 'error'
               : undefined
+        }
+        help={classificationMethod === 'kmeans'
+          ? locale.numberOfRulesViaKmeans
+          : numberOfRules < minNrClasses
+            // eslint-disable-next-line max-len
+            ? `${locale.colorRampMinClassesWarningPre} ${minNrClasses} ${locale.colorRampMinClassesWarningPost}`
+            : undefined
+        }
```
KaiVolland, line 234: I know this is not part of thos PR but would be nice if we would have this human readable.

src/Util/FormItemUtil.spec.ts
```diff
@@ -0,0 +1,38 @@
...
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+import { getFormItemConfig } from './FormItemUtil';
+
+describe('FormItemUtil', () => {
```
KaiVolland, line 31: This test is not necessary.

KaiVolland (approved):

PR #2480, opened 2024-06-17
feat: produce esm build

hwbllmnn:
## Description

This switches the type of the module to 'module' and adapts the typescript config to produce an esm build.

- [x] Yes
- [ ] No

@geostyler/devs Please review.


styleguide/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
jansule, line 1: this file can probably be removed, right?

jansule (commented):
This looks like a breaking change, right? So adjusting the commit message would be great

styleguide/index.html
hwbllmnn: Done

PR #2484, opened 2024-06-21
Migrate tests to use vitest

hwbllmnn:
## Description

See title.

- [x] Yes
- [ ] No

@geostyler/devs Please review.


jest/setup.js
KaiVolland, line 1: Maybe the folder should be renamed to `test/setup.js` or similiar.

vitest.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from 'vitest/config';
+
+export default defineConfig({
+  test: {
+    server: {
+      deps: {
+        inline: true
+      }
+    },
+    setupFiles: ['./jest/setup.js'],
```
KaiVolland, line 10:
```suggestion
    setupFiles: ['./test/setup.js'],
```

hwbllmnn: done

KaiVolland (approved):

PR #2502, opened 2024-10-09
Fix color picker

### LINKED ISSUE #2501: Issue of the ColorPicker
### Description

Hello,
When integrating the GeoStyler to manipulate the style of drawing tools, the transfer of the fill colour does not work as expected.
After changing the colour, an rgb value is displayed in the colour bar, which is also returned, instead of a hex value. In addition, the bar for changing the transparency (alpha value) is available in the colour picker.
The same behaviour occurs with the stroke colour, but here the transfer of the colour works.
Ant version 5.21.2 is used in the project.

### Expected behaviour

Chosen color in the ColorPicker should be displayed correctly for each drawn feature and the transparency bar should not be visible

### How to reproduce

Include GeoStyler in a project with ant-version 5.21.2

### Possible Solution

_No response_

### Additional Context

_No response_

KaiVolland:
## Description
This fixes some issues with the ColorPicker: Compare #2501 

- allow RGB/HSB mode
- hide alpha inputs via API not via CSS
- ensure value of onChange is always in hex format

It also updates @ant-design/icons

### Preview

geostyler_colorfield

Fixes #2501 

- [ ] Yes
- [x] No

### DIFF
--- a/src/Component/Symbolizer/Field/ColorField/ColorField.example.md
+++ b/src/Component/Symbolizer/Field/ColorField/ColorField.example.md
@@ -36,12 +36,11 @@ import { ColorField } from 'geostyler';
 
 const ColorFieldExample = () => {
 
-  const [value, setValue] = useState();
+  const [value, setValue] = useState("#013370");
 
   return (
     <ColorField
       value={value}
-      defaultValue="#013370"
       label="Color"
       onChange={setValue}
     />
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "15.0.0-beta.5",
       "license": "BSD-2-Clause",
       "dependencies": {
-        "@ant-design/icons": "^5.3.7",
+        "@ant-design/icons": "^5.5.1",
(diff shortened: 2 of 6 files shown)


src/Component/Symbolizer/Field/ColorField/ColorField.example.md
```diff
@@ -36,12 +36,11 @@ import { ColorField } from 'geostyler';
 
 const ColorFieldExample = () => {
 
-  const [value, setValue] = useState();
+  const [value, setValue] = useState("#013370");
```
jansule, line 39: Will this lead to the same behavior as before?

KaiVolland: This is just the example code. The behaviour is visible in the gif.

jansule: Good point :grinning:

jansule (approved):
Very nice!

PR #2529, opened 2024-11-28
Add placeholder for renderer with functions

### LINKED ISSUE #2526: Symbol Preview When Using Expressions
### Description

When using expressions, we currently do not provide a symbol preview, which might confuse users and it requires quite some mental load, when creating complex styles with multiple symbolizers. We should either find a way to display different outcomes of the used expressions (if possible), or at least show an indication about an expression being used. Maybe, displaying the basic symbol type together with the expressions icon might be a good start.

### Example

_No response_

### Alternatives

_No response_

### Additional Context

_No response_

KaiVolland:
## Description

This adds a placeholder icon to the ol style when a GeoStyler symbolizer with functions is used. An additional :information_source: icon is shown in the bottom right to explain why.

It closes #2526 

## Preview

image

- [ ] Yes
- [x] No


src/Component/Renderer/OlRenderer/placeholder.ts
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable max-len */
+/**
+ * This is the base64 datauri representation of the placeholder image.
+ * The "FunctionOutlined" icon from antd was the source. Compare ./placeholder.svg
```
marcjansen, line 4:
```suggestion
 * The "FunctionOutlined" icon from antd (MIT-licensed) was the source. Compare ./placeholder.svg
```

See https://github.com/ant-design/ant-design-icons I think it makes sense to mention the license.

marcjansen (approved):
Thanks, looks great. One suggestion to mention the license of the source of the icon.

PR #2531, opened 2024-11-28
Improve props composition for FontPicker

### LINKED ISSUE #2405: Allow setting list of fonts to FontPicker
### Description

The FontPicker currently works with a hardcoded list of supported fonts.

### Expected behaviour

It should be able to pass a list of supported fonts via the composition context or any other mechanism, to enable users to select other fonts than the predefined ones.

### How to reproduce

--

### Possible Solution

_No response_

### Additional Context

_No response_

KaiVolland:
BREAKING CHANGE: The FontPickers value name changed from `font` to `value`

## Description

This improves the props of the `FontPicker` componet.

It is now possible to provide a `defaultValue` and list of `fonts` via GeoStyler context.
This closes #2405 

:warning: Strictly speaking, this is a breaking change because the props of a component have changed. Even if this component is certainly only used internally.

- [x] Yes
- [ ] No

### DIFF
--- a/src/Component/Symbolizer/Field/FontPicker/FontPicker.example.md
+++ b/src/Component/Symbolizer/Field/FontPicker/FontPicker.example.md
@@ -39,27 +39,26 @@ class FontPickerExample extends React.Component {
     super(props);
 
     this.state = {
-      font: ['arial']
+      value: ['arial']
     };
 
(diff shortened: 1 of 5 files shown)


jansule (approved):
Very nice!

PR #2532, opened 2024-11-28
Remove unused props

### LINKED ISSUE #2466: Remove Unused Properties in Rules Component
`<Rules>` provides two properties `showAmount`, `showDuplicates` which are not being used by the component anymore (now handled via composition context). These should be removed. Since this introduces a breaking change, we should wait for the code sprint to do so.

 https://github.com/geostyler/geostyler/blob/141238038a867f0f59aad1b8208093a7279997bb/src/Component/Rules/Rules.tsx#L63-L67

KaiVolland:
## Description

This removes the unused `showAmount` and `showDuplicates` props from the `Rules` component.
The visibility of these fields can be configured via `Rule` in the GeoStyler composition context.

This closes #2466 

- [x] Yes
- [ ] No

### DIFF
--- a/src/Component/Rules/Rules.tsx
+++ b/src/Component/Rules/Rules.tsx
@@ -60,10 +60,6 @@ export interface RulesComposableProps {
 }
 
 export interface RulesInternalProps {
-  /** Display the number of features that match a rule */
-  showAmount?: boolean;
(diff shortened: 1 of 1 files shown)


jansule (approved):
Very nice!

PR #2536, opened 2025-01-03
chore: Pin chroma-js to 2.4.2

ploth:
## Description

* chroma-js introduced breaking changes in newer versions after `2.4.2` which is currently specified in the package.json (^2.4.2).
* I reported the issue to chroma-js here: https://github.com/gka/chroma.js/issues/367 
* Projects depending on geostyler cannot be build currently as the `^` is allowing newer minor versions of chroma-js (despite using e.g. version resolution from yarn: https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)
* Renovate had also issues with it: https://github.com/geostyler/geostyler/pull/2519

- [ ] Yes
- [x] No

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22,7 +22,7 @@
         "@types/lodash": "^4.17.5",
         "@ungap/url-search-params": "^0.2.2",
         "antd": "^5.18.3",
-        "chroma-js": "^2.4.2",
+        "chroma-js": "2.4.2",
         "color": "^4.2.3",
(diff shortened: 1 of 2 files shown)


KaiVolland (approved):
Thanks. Good catch.

KaiVolland:
Just clicked your link after merging. :see_no_evil: If it's fixed with chroma.js >2.5 upgrading would be better i guess.

Issue #2576, opened 2025-06-03
Apply or cancel modifications made in modals


slafayIGN:
### Description

When using the RuleTable, there are 2 modals : 
* the symbolizer editor
* the filter editor

In each of them you can make edits that are instantly applied to the style.

The idea is to apply the modification only to a style in the context of the modal and apply it only when clicking on an explicit "Apply" or "Save" (in french "Enregistrer") button in the modal.

Exiting the modal would be a cancel action and no modification would be applied on the style object.

### Additional Context

This interaction should be available only in modals.
It should not be used in the symbolizer editor in the Card Layout. It should keep a "real time" behaviour.

PR #2580, opened 2025-06-04
apply or cancel modifications made in modals #2576

pprev94:
Apply or cancel modifications made in modals #2576

Normally n breaking changes


KaiVolland (approved):
Nice. Just some remarks regarding translations.

PR #2583, opened 2025-06-04
Feat/rules drag and drop

mircmo:
## Description

This PR adds drag and drop functionality to the rules table.  
The reason why the implementation may look needlessly complex, is that I tried to avoid any changes to the interfaces and/or exported components.

https://github.com/user-attachments/assets/b21538aa-9ed3-4a47-b420-2bf8d3c048fe

- [ ] Yes
- [x] No


src/Component/RuleTable/RuleTable.tsx
```diff
@@ -85,8 +95,8 @@ export interface RuleTableInternalProps {
   onRulesChange?: (rules: GsRule[]) => void;
   /** The callback function that is triggered when the selection changes */
   onSelectionChange?: (selectedRowKeys: string[], selectedRows: any[]) => void;
-  onCloneRule?: (rule: RuleRecord) => void;
-  onRemoveRule?: (rule: RuleRecord) => void;
+  onCloneRule?: (index: number) => void;
+  onRemoveRule?: (index: number) => void;
```
jansule, line 99: Just to keep in mind that this is a breaking change. While this is not a problem, we should provide information about it in the final commit message.

jansule (approved):
Very nice! @mircmo

slafayIGN (approved):
Looks good to me. That's a huge work. Thanks !

I think a second reviewer is necessary.

src/Component/RuleTable/RuleTable.tsx
mircmo: kind of, but this feature was added yesterday by arnest, so it's not used yet

PR #2600, opened 2025-06-11
Feat/min max denominators

pprev94:
## Description
- The two fields min and max scale denominators are in the same component wich can be either two InputNumber or two Select depending on scaleDenominators wich is defined or not (data.scaleDenominators in the context). 
**geostyler-data must be released**

- There's validation for InputNumber (max must be greater than min)
image

- For Selects, options are controlled (min is always less then max)
image

## WARNING
**It depends on geostyler-data wich must be released**

### DIFF
--- a/src/Component/ExpressionInput/NumberExpressionInput/NumberExpressionInput.tsx
+++ b/src/Component/ExpressionInput/NumberExpressionInput/NumberExpressionInput.tsx
@@ -44,9 +44,15 @@ import { FunctionOutlined } from '@ant-design/icons';
 
 import './NumberExpressionInput.css';
 
+type SliderProps = {
+  min?: number;
+  max?: number;
+  step?: number;
+};
 export interface NumberExpressionInputProps {
   className?: string;
   slider?: boolean;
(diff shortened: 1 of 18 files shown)


src/Component/ExpressionInput/NumberExpressionInput/NumberExpressionInput.tsx
```diff
@@ -92,6 +92,9 @@ export const NumberExpressionInput: React.FC<NumberExpressionInputProps> = ({
       {slider ? (
         <div className={'slider-wrapper'}>
           <Slider
+            min={0}
+            max={1}
+            step={0.01}
             value={inputValue}
             range={false}
             onChange={(val) => {
```
KaiVolland, line 100: It would be useful to have an `sliderProps` porperty on the NumberExpressionInput analog to the `inputProps` so you can pass props to the slider.

KaiVolland (approved):
Very nice. Just one minor remark but its not blocking.

KaiVolland:
This PR should also include the update of `geostyler-data` to version 1.1.0.

PR #2616, opened 2025-06-19
fix: height renderer in symbolizer editor

cbrousseau1:
## Description
correction of the height of renderer in symbolizer editor

- [ ] Yes
- [x] No


KaiVolland:
Could you provide a before/after screenshot and/or a short explanation why this is needed?

cbrousseau1:
the rendering is the same but it's extended to general use, because until now this height was only defined in the case of the demo example. As a result, when we used it outside the example, we had a rendering with a very small height because the height of the example didn't apply.

KaiVolland (approved):

Issue #2618, opened 2025-06-20
Add disableName in StyleComposableProps


pprev94:
### Description

The problematic :

PR #2623, opened 2025-06-24
feat: add option to disable name #2618

pprev94:
## Description

We have an use case where we don't want to let the user change the name of the style and handle it ourselves (because the name of the style must correspond to the name of the data). 
I've added disableName (optional) in StyleComposableProps, false by default in Style
This option disable or not the name input of the style.

### DIFF
--- a/src/Component/NameField/NameField.tsx
+++ b/src/Component/NameField/NameField.tsx
@@ -38,8 +38,6 @@ export interface NameFieldProps {
   placeholder?: string;
   /** The value to display in input field */
   value?: string | undefined;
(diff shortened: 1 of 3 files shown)


ocruze (approved):

pprev94:
OK and implement like this :

nameField?: {
    visibility?: boolean;
}

Do you agree ?

abohaboh195 (approved):

jansule (approved):
Very nice! Thanks!

abohaboh195 (approved):

PR #2649, opened 2025-09-26
Explicitly remove outputFormat parameter from DescribeFeatureType request

mki-c2c:
A Geostyler server will not integrate correctly with a standalone instance of geoserver.

This is because by default geoserver will return a json style description which cannot be parser by geostyler.

Via this fix, no output format is requested explicitly, then geoserver responds with valid XML

- [ ] Yes
- [ ] No
- [x] I am unsure (no worries, we'll find out)


mki-c2c:
Hello @jansule 
I was not sure about it. I first submitted it for personal use, and noticed, I did not know the CI very well.

It seems to be stable though, so feel free to review.

I also fixed the commit message for CI

PR #2663, opened 2025-12-16
Fix format check

hwbllmnn:
## Description

Fixes the file format check to work in the browser build.

- [ ] Yes
- [x] No


src/Component/CodeEditor/CodeEditor.tsx
```diff
@@ -115,19 +114,19 @@ const getFileFormat = (parser: StyleParser): FileFormat => {
     mimeType: 'text/plain',
   };
 
-  if (parser instanceof SldStyleParser){
+  if (parser && parser.title === 'SLD Style Parser'){
```
jansule: Would it be better to check against the static field `title` of each parser in order to determine which one should be used, instead of using hard coded strings?

So something like

```
if (parser && parser.title === SldStyleParser.title)
```

hwbllmnn: Yes, changed it

PR #2720, opened 2026-03-02
Add support for between filter

### LINKED ISSUE #2659: Bug: Missing handling of <= x <= (between) operator in handleSimpleFilter
Hi,

we noticed that the `handleSimpleFilter` method in FilterUtil.ts does not implement handler for the _between_ operator.
Currently the method supports comparison operators such as '==', '*=', '!=', '<', '<=', '>', '>=', but there is no logic for expressions like '<=x<=', where a value is constrained on both sides.
A the moment, using this operator causes the method to throw an Error.

https://github.com/geostyler/geostyler/blob/b3cd7e08443a19b42d41ef4cdb6dc95997361586/src/Util/FilterUtil.ts#L104

KaiVolland:
Introduce support for the between filter in the filtering utility and update the FilterOverview example to demonstrate its usage.

## Preview

image

Fixes #2659

### DIFF
--- a/src/Util/FilterUtil.spec.ts
+++ b/src/Util/FilterUtil.spec.ts
@@ -38,6 +38,233 @@ describe('FilterUtil', () => {
     filter = TestUtil.getDummyGsFilter();
   });
 
+  describe('handleSimpleFilter', () => {
+    const dummyData = TestUtil.getComplexGsDummyData();
+    const feature = dummyData.exampleFeatures.features[0];
+
+    describe('equality operator (==)', () => {
+      it('returns true when values are equal', () => {
+        feature.properties.name = 'Berlin';
+        const simpleFilter: Filter = ['==', 'name', 'Berlin'];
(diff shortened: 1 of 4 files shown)


src/Util/FilterUtil.spec.ts
```diff
@@ -38,6 +38,233 @@ describe('FilterUtil', () => {
...
   });
 
+  describe('handleSimpleFilter', () => {
+    const dummyData = TestUtil.getComplexGsDummyData();
+    const feature = dummyData.exampleFeatures.features[0];
+
+    describe('equality operator (==)', () => {
+      it('returns true when values are equal', () => {
+        feature.properties.name = 'Berlin';
+        const filter: Filter = ['==', 'name', 'Berlin'];
```
jansule: you might want to adjust the variable name here. otherwise this looks good!

jansule (approved):
Very nice!

Issue #2739, opened 2026-04-20
ColorPreview in RuleGenerator seems broken


jansule:
### Description

The ColorPreview in the RuleGenerator component seems to be broken. It does not show anything

<img width="820" height="835" alt="Image" src="https://github.com/user-attachments/assets/9fccd302-2572-4fb5-ba80-d7d51f43aa9d" />

### Expected behaviour

It should show something (not sure what exactly, to be perfectly honest)

### How to reproduce

1. go to https://demo.geostyler.org/
2. upload any dataset, e.g. a geojson
3. click `classification`
4. fill out all fields
5. color preview does not show anything

KaiVolland:
Fixed with #2740

PR #2740, opened 2026-04-21
Add min-width to ColorPreview

KaiVolland:
## Description

This fixes the broken ColorPreview #2739 by adding a min-width and min-height.

- [ ] Yes
- [x] No


jansule (approved):
Very nice!

PR #2741, opened 2026-04-21
Allow autoselect fields for numeric properties

hwbllmnn:
## Description

On first glance, the autoselect feature does not make sense for numberic fields. However, often numeric codes are used with codelists for eg. categories etc., where this could be a useful feature.

- [ ] Yes
- [x] No


jansule (approved):
Very nice!

PR #2748, opened 2026-04-30
fix: adjust incorrect import path in GeoStylerContext

jansule:
## Description

This fixes an incorrect import path.

- [ ] Yes
- [x] No


KaiVolland (approved):
I think we should add path-aliases to the typescript and vite configs so we can use absolute import pathes.

jansule:
> I think we should add path-aliases to the typescript and vite configs so we can use absolute import pathes.

Might be a nice improvement, if that reduces the risk of such errors. Can you create an issues for that?

